PR #246, opened 2022-09-25
Update README.md

tur11ng:
Fix typo.


xusheng6:
Thx for the PR! I slightly modified your texts and now it is in https://github.com/Vector35/debugger/commit/047c977d831ff854bb1332c564d0879c46a1ca34. It will be available once we push to the dev the next time

PR #332, opened 2022-11-02
add support for showing end address of modules

### LINKED ISSUE #241: Support showing end address of modules

op2786:
Tried to fix #241.

I think renaming "address" to "base" and "end address" to "end" would be good idea. I didn't do it because I wasn't sure if you would like it.


xusheng6:
This causes the full path column to vanish. Otherwise, it looks good to me.

xusheng6:
Merged into https://github.com/Vector35/debugger/commit/b045c2a56ee98fd53aab0c081195b0b50d9ba50f

PR #333, opened 2022-11-02
added jump to ip action to debugger menu

### LINKED ISSUE #312: Add an action for "navigate to PC"
This would supplement #229 nicely and be very easy to add

op2786:
fix #312.

I'm not sure if I should add "Jump to IP" in right click debugger menu. I find it hard to use because I need to click four times. If we need so, please let me know.


psifertex:
I think just having the action is great. Agree that a buried right click doesn't really help. Only advantage is if maybe it helps someone discover that it exists but not sure it's worth cluttering the UI for that. 

Thanks for the PR!

PR #336, opened 2022-11-03
added set ip action

### LINKED ISSUE #192: Add support for overriding the instruction pointer

op2786:
fixes #192

### DIFF
--- a/api/debuggerapi.h
+++ b/api/debuggerapi.h
@@ -504,6 +504,7 @@ namespace BinaryNinjaDebuggerAPI
 
 		uint64_t IP();
 		uint64_t GetLastIP();
+		bool SetIP(uint64_t address);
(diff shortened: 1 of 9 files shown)


xusheng6:
Thx for the PR! It works fine in most cases, except for a small issue with LLDB on macOS. That LLDB seems to report the old program counter even if we have already modified it. This causes the UI IP highlight to be wrong. 

Also, I am weighing popping up a dialog for the user to type in the new instruction pointer value, and has its default value to be the current address. This way, the user can confirm his action.

I will take over and commit any changes needed on top of it, and then push to dev.

xusheng6:
This is merged into https://github.com/Vector35/debugger/commit/d0e25a9bec8d61064196318f632228a67bfe57d3

PR #337, opened 2022-11-03
added restoring debug adapter settings from database

### LINKED ISSUE #313: Debug adapter settings are not serialized into the database
Debug adapter settings, including the executable path, working directory, and most noticeably, the command line arguments are not serialized into the database. They are lost once we save and re-open the database.

op2786:
fixes #313.

Also fixed an incorrent type check at https://github.com/Vector35/debugger/blob/88518015c56ce2599b957b5547b2db992aacd64c/core/debuggerstate.cpp#L728

### DIFF
--- a/core/debuggerstate.cpp
+++ b/core/debuggerstate.cpp
@@ -725,7 +725,7 @@ DebuggerState::DebuggerState(BinaryViewRef data, DebuggerController* controller)
         m_adapterType = metadata->GetString();
 
     metadata = m_controller->GetData()->QueryMetadata("debugger.terminal_emulator");
-    if (metadata && metadata->IsUnsignedInteger())
+    if (metadata && metadata->IsBoolean())
         m_requestTerminalEmulator = metadata->GetBoolean();
     else
         m_requestTerminalEmulator = false;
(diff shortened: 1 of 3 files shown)


xusheng6:
This PR is a little bit problematic, as the saved adapter settings are only restored if the user opens the adapter settings dialog. However, currently, they can launch the target directly, without needing to open the adapter settings dialog. 

The un-serialization should be done as part of DebuggerState initialization. You can have a look at when `DebuggerBreakpoints::UnserializedMetadata()` is called, do deal with these adapter settings at a similar time. 

Also, I notice there is also a problem in the way that I serialize these information. They are only stored if they are changed from the adapter settings dialog. They can be changed from the API as well, which will not be saved into the database. Again, to deal with this, you can have a look at how I serialize breakpoints.

If you would like to, please make the above changes and I will review the PR again.

xusheng6:
Issue #313 is fixed in https://github.com/Vector35/debugger/commit/b50967d260b0bad548f3c2466cf5f6f74ce26f5c. Thanks again and sorry for the inconvenience!

PR #356, opened 2022-11-09
various improvements for thread frames view

### LINKED ISSUE #325: Support navigating to pc, sp, fp in the stack trace
Currently, double-clicking an item in the stack trace navigates to the program counter of the frame. We should also offer to navigate to the stack pointer and frame pointer, maybe via a context menu.

op2786:
* fixed a bug in dbgeng adapter which was frame index did not set when enumerating them.
* double clicking on PC, SP and FP now navigates to address, fixes #325.
* converted listview to tableview as we discussed in slack. But I did not use filtered view, please let me know if I need to.

Here is a preview:

frames

Btw, is there a problem at getting function names? I couldn't be able to get any function names in my tests.


xusheng6:
Merged into c87eafc185ac906e9e373fafa22fb4353ae583fc

PR #368, opened 2022-11-15
fix register name/value column width not wide enough in some situations

### LINKED ISSUE #269: Default register name column width is not wide enough for register names
image

op2786:
fixes #269


xusheng6 (changes requested):
I suggest you put these lines into a function and call it. iirc there are other places where the column width is updated, and we should also replace it with calling the function

xusheng6:
LGTM, I will merge it later

xusheng6:
Added in https://github.com/Vector35/debugger/commit/77c433d66199acdf6492c59c860c3fa89d94d5e2

PR #371, opened 2022-11-16
added option for checking dbgeng dlls

### LINKED ISSUE #343: Warn users that the debugger does not work on Windows if the dbgeng DLLs are loaded before starting debugging
If someone (presumably anti-virus), loads those dbgeng DLLs into BN before we start debugging, the debugging will fail. That is because the dbgeng DLLs in system32 will be loaded, and they are incomplatible with the dbgeng that our debugger uses.

op2786:
fixes #343

### DIFF
--- a/core/adapters/dbgengadapter.cpp
+++ b/core/adapters/dbgengadapter.cpp
@@ -104,48 +104,80 @@ std::string DbgEngAdapter::GetDbgEngPath(const std::string& arch)
 
 bool DbgEngAdapter::LoadDngEngLibraries()
 {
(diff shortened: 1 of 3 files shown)


xusheng6:
Merged in https://github.com/Vector35/debugger/commit/972745e355ce518de5a36f0607b8b57475a5426b

xusheng6:
@op2786 Because you accidentally checked in the binaryninja-api submodule, I have no way but cherry-pick those code changes from your commit and then push it. And as an unintended side-effect, the commit now becomes "mine". I hope you are fine with it

Issue #374, opened 2022-11-18
Support for listing threads


op2786:
Currently, we can see thread list in stack trace widget but it can be improved in different ways. For example:

* We can have another tab in the debugger sidebar like registers /breakpoints
* We can have another widget in global area
* We can modify the thread frames widget into tree view and make it list threads as first level, and thread frames of the thread if expanded

xusheng6:
Also fixed in 5372a0f1411afd3ca4f22f635a77c72b0afe812f

PR #377, opened 2022-11-22
Thread suspend resume

xusheng6:


xusheng6:
image

I think we should avoid duplicating these `0xae36 @ 0x5555555551e0`, only showing it on the first line suffices

xusheng6:
Also here, https://github.com/Vector35/debugger/pull/377/files#diff-341368eee4931026af7dfb4eb9e814f7392695f0651f1ce8456becdcdc5d3770R641-R656, for dbgeng, I think we really should do freeze/unfreeze threads, rather than suspend/resume thread, whose semantics can be a bit confusing. The Resume()/Suspend() in SBThread in LLDB is also the same as freezing/unfreezing in dbgeng.

Also, we need a status to indicate whether a thread is frozen or not. Currently, if we click the suspend/resume thread, there is no UI feedback. This is quite a bit work since you will need to add a new field in the DebugThread struct, and read its value from the dbgeng back-end.

xusheng6:
Additionally, we need something like a solo thread which is quite common in vfx software. Basically, if we solo a thread, all other threads are frozen and only the solo thread will run. This can be done as a UI feature, as long as we can freeze/unfreeze one thread, this should be easy to implement.

xusheng6:
Otherwise, the appearance look good and once we get these problems fixed, this can be fixed and it will make the threads widget much more usable. Thanks for the contribution!

xusheng6:
Here, https://github.com/Vector35/debugger/pull/377/files#diff-7cd18de9cec6256b4d4d62a8a4c82dcfb979c68d551a6d5b697313d5499563d9R460, we should use `bool Suspend(SBError &error);` overload to check for any errors

xusheng6:
https://github.com/Vector35/debugger/pull/377/files#diff-690ea7c3b09a7fe5931676f9498f62b8dd1a677e15ff314e1c7b300be11e6439R379-R383, is there any reason that we cannot resume/suspend the current active thread?

xusheng6:
This PR addresses https://github.com/Vector35/debugger/issues/374

PR #378, opened 2022-11-22
fix debugger modules last column does not stretch

op2786:


xusheng6:
Available in 8ef8c63211101119251b5f72e06329e66e46352a

xusheng6:
I merged this PR despite https://github.com/Vector35/debugger/issues/381, since it makes things look better in most cases.

PR #379, opened 2022-11-22
executed clang-format in order to fix some indent issues etc.

op2786:
actually I didn't liked some of the formatting clang-format done, but generally I obeyed .clang-format file. I've used the .clang-format from binaryninja-api repo.

I've checked files by eye, and fixed some ugly comments, function calls parameters that dropped next line when it's not necessary etc.


xusheng6:
Available in b4766867bb557c1b2d4ff515c0c10f21e549f711

PR #398, opened 2022-12-12
thread suspend resume support

### LINKED ISSUE #373: Support for suspending /resuming individual threads

op2786:
fixes #373, #374

This PR adds support for thread suspend /resume for both LLDB and dbgeng backends. It also makes stack trace view a treeview in order to show both threads and their frames.

NOTE: I did not remove some log output to make it easier to understand that the code is working correctly. I guess we should remove them before merging.

Preview:

preview

### DIFF
--- a/api/debuggerapi.h
+++ b/api/debuggerapi.h
@@ -276,6 +276,7 @@ namespace BinaryNinjaDebuggerAPI {
 	{
 		std::uint32_t m_tid {};
 		std::uintptr_t m_rip {};
+		bool m_isFrozen {};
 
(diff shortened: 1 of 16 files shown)


xusheng6:
It looks pretty good!  Thanks for your hard work.

I think you will need to make the following changes before I merge it to dev:

1. Remove all debug log messages
2. Always expand the current thread when we update the thread list. We will deal with keeping the expanded/collapsed state in the future
3. Squash all commits, rebase on top of the latest dev.

Once you are done with these, please let me know and I will merge it!

PR #399, opened 2022-12-12
better register support for python api

### LINKED ISSUE #349: Better support for registers in Python API
It would be really nice if we can slice the registers like `dbg.regs['rax']`, and it would be even better if we can change its value by `dbg.regs['rax'] = 0x1234`.

op2786:
fixes #349

adds basic support for slicing registers


xusheng6:
Please remove the binaryininja-api submodule

op2786:
oh I should have moved that submodule to another folder already...

xusheng6:
Merged into 573d21e693d6fb27bb6afe7e266f93a4b3e69cca

PR #541, opened 2024-02-05
Draft TTD API and UI

JohnathanKG:
This PR is a draft for declaring adapters as TTD-compatible, buttons to navigate from the debugger GUI, a python API, and a unit test for the TTD API
Currently it is lacking fully proper documentation in the Python API, and custom icons do not work.

### DIFF
--- a/api/debuggerapi.h
+++ b/api/debuggerapi.h
@@ -509,7 +509,8 @@ namespace BinaryNinjaDebuggerAPI {
 		bool StepOver(BNFunctionGraphType il = NormalFunctionGraph);
 		bool StepOverReverse(BNFunctionGraphType il = NormalFunctionGraph);
 		bool StepReturn();
-		
+		bool StepReturnReverse();
+
 		bool RunTo(uint64_t remoteAddresses);
 		bool RunTo(const std::vector<uint64_t>& remoteAddresses);
 		void Pause();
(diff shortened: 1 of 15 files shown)


xusheng6:
Here are a few quick comments so you can start working on it:

1. We should also add support for "step return reverse". The DbgEng does not offer constant like `DEBUG_STATUS_REVERSE_STEP_OVER` for reverse step return. But you can look at how step return (forward) is implemented, basically it just sends a "gu" command (https://github.com/Vector35/debugger/blob/318e61219ced0815351eea16c6388d3e1372e6f7/core/adapters/dbgengadapter.cpp#L1194-L1201). And to step return reverse, you need to send a "g-u" command
2. For the icons, just use the same ones as the forward operations. We will create new ones soon
3. Actions like "Step Over" are connected to keyboard shortcuts. We should do the same for the newly add actions. You can see how I was doing it here: https://github.com/Vector35/debugger/blob/471917726e646d28390d777902e5e846ef01ec7c/ui/ui.cpp#L452. I think we should add a Shift to the keybindings for the forward actions. For example, step over is F8, then we should make step over reverse shift+F8

xusheng6:
@JohnathanKG could you please remove the TTD related unit test (for now)? I tested and the trace generated on your box does not work properly on mine. And it probably also would not work on our infra.

xusheng6:
Merged into https://github.com/Vector35/debugger/commit/7cd18122ad881e17a64d9ef63ceb2c56899e21b6. Thanks for the contribution!

Issue #645, opened 2024-10-31
Make "F4" key binding default to "Run To Here"


blue-devil:
Hello folks;

One of the best things in a debugger is `Run To Here` debug command. That makes debugging faster.
But on Binary Ninja we don't have a key binding of it. Generally `F4` is used for that:

| App Name | Operation Name | Key Binding|
| --- | ---- | --- |
| IDA Pro | `Run to cursor` | `F4` |
| x64dbg | `Run Until Selection` | `F4` |
| Binary Ninja | `Run To Here` | not set |

Quick and personal workaround:
`Ctrl + P (Command Palette) > Key Bindings` and then set `F4` for key binding.

My wish is; can you consider making `F4` keybinding default for `Run To Here` debug command?

Thank you.

xusheng6:
Sounds like a good idea. I did not even know other tools do it

xusheng6:
Added in https://github.com/Vector35/debugger/commit/7a5b384cca01e6064283272a5efe70922d4a7c63

PR #663, opened 2024-11-18
Added F4 keybinding for action Run To Here.

1ikeadragon:
Ref: #645


xusheng6:
Thx a lot! @1ikeadragon It will be available in the dev builds soon

PR #772, opened 2025-08-15
Add dependency tracking to the Python code generation

bdash:
This ensures that the Python source files are only generated and copied into the output directory if inputs have changed, rather than being done unconditionally.

Builds on the similar change made in https://github.com/Vector35/binaryninja-api/commit/3e2ace5e8d127d7995a94f1b296cad9152cb7274


xusheng6:
LGTM, thx! @bdash

PR #873, opened 2025-10-10
Add active_pid property to get debugged process ID

3rdit:
Added `active_pid` property to get the PID of the process that is currently being debugged. Returns 0 if unavailable or nothing is yet attached. 

Implementation is added for every adapter, for GDB it's cached at connection time (as then we know the thread id equals the process pid).

### DIFF
--- a/core/adapters/queuedadapter.cpp
+++ b/core/adapters/queuedadapter.cpp
@@ -568,6 +568,22 @@ uint64_t QueuedAdapter::GetStackPointer()
 }
 
 
+std::uint32_t QueuedAdapter::GetActivePID()
+{
+    std::unique_lock<std::mutex> lock(m_queueMutex);
+
+    std::uint32_t ret;
+    Semaphore sem;
+    m_queue.push([&]{
+        ret = m_adapter->GetActivePID();
+        sem.Release();
+    });
+    lock.unlock();
+    sem.Wait();
+    return ret;
+}
+
+
 bool QueuedAdapter::SupportFeature(DebugAdapterCapacity feature)
 {
     return m_adapter->SupportFeature(feature);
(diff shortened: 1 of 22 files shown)


core/adapters/queuedadapter.cpp
```diff
@@ -568,6 +568,22 @@ uint64_t QueuedAdapter::GetStackPointer()
 }
 
 
+std::uint32_t QueuedAdapter::GetActivePID()
```
xusheng6: This is no need to change anything in queedadapter.cpp, it is a legacy file that is no longer used

core/debuggercontroller.cpp
```diff
@@ -2107,6 +2107,15 @@ uint32_t DebuggerController::GetExitCode()
 }
 
 
+uint32_t DebuggerController::GetActivePID()
+{
+	std::lock_guard<std::mutex> lock(m_adapterMutex);
```
xusheng6: No need to take the lock here

xusheng6 (commented):
I made two small comments. I will take care of them and then do some functionality tests

One thing that I am not sure about is that the debugger normally uses the DebuggerState object to cache the information about the target, e.g., the registers, threads, etc. And it is updated whenever necessary, e.g., when the target stops. This PR does not follow that pattern. However, I have not yet decided for things like the PID, whether we want it to be cached at all -- because it does not change very often, and it is also not frequently used

xusheng6:
@3rdit thanks for the contribution!

xusheng6:
@3rdit I am actually still undecided regarding whether we should just forward the request to the adapter or cache it in DebuggerState. I just went ahead and merged it to meet our feature freeze deadline. I will reconsider it after the release, and may file a new issue to cache it in DebuggerState. If that is the case, and you are still interestedin contributing, I will ping you at that time!

3rdit:
Thanks for the context, @xusheng6. If you do decide to revisit that caching approach and need additional help, give me a ping.

PR #936, opened 2025-12-06
Add module-relative address resolution for breakpoints

### LINKED ISSUE #886: Resolve module filenames to their base address when debugging
This request is focused on saving breakpoints but has other benefits. Ultimately, treat copy/pasted breakpoint addresses such that they can be reused.

Thanks to a recent change, we are now able to copy the Breakpoint table view out of Binary Ninja. This is good for basic documentation but they cannot be reused without additional effort.

Current output:
filename.exe + 0x0111f

Current workflow:
While debugging:
Go to module list and find RVA of filename.exe
Replace filename.exe with RVA in external text editor
Go to that address, like "0x7ff770f60000 + 0x0111f", and set a new breakpoint.

Feature request:
Allow address resolution (go to) to support the filename of any loaded module. If "filename.exe" is entered, then find the base address of that file and use it for relative lookup. If filename.exe is not seen in module list, then fail gracefully as if a wrong value was entered (like it currently does)

Requested workflow:
Copy "filename.exe + 0x111f" from breakpoint table
Rerun application, which now has a new base address
Go to address "filename.exe + 0x111f" and set a breakpoint

3rdit:
The breakpoint addresses copied from the table (e.g., `module + 0x1234`) can now be pasted and used directly in the add breakpoint dialog.

The module names are registered in the expression parser when the debugger stops, resolving to their base addresses automatically.

This PR closes #886.

### DIFF
--- a/api/debuggerapi.h
+++ b/api/debuggerapi.h
@@ -638,6 +638,7 @@ namespace BinaryNinjaDebuggerAPI {
 		bool ResumeThread(std::uint32_t tid);
 
 		std::vector<DebugModule> GetModules();
+		bool ParseModuleRelativeAddress(const std::string& input, uint64_t& result);
 		std::vector<DebugRegister> GetRegisters();
 		intx::uint512 GetRegisterValue(const std::string& name);
 		bool SetRegisterValue(const std::string& name, const intx::uint512& value);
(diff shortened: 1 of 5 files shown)


xusheng6:
Hi @3rdit thanks for the PR. I have cherry-picked the module name change into https://github.com/Vector35/debugger/commit/f81b54ae76d7b5db7f8a5bb889a98448f6a04bdd. 

For the breakpoint parsing change, I would like to reject it for now. I feel like a more proper way to implement it is to do a breakpoint export and import, from where the custom parsing can be done

PR #939, opened 2025-12-21
Fix many of the warnings that show up when compiling with GCC 15.2

bdash:
Many of the warnings were related to misleading indentation due to code within a function using a mixture of tabs and spaces. This could lead to code after single-line if statements potentially appearing to be indented to the level of the if statement body, depending on how the editor displayed tabs vs spaces.


xusheng6:
@bdash thank a lot!

Issue #53, opened 2021-12-16
Support hardware breakpoint


xusheng6:

xusheng6:
I think it is time to get this done. Previously I wrongly think LLDB does not support hardware execute breakpoint, but I am wrong, you can set such a breakpoint like `break set -H -name main`, and hardware read/write breakpoint can be set via the `watchpoint` command

xusheng6:
Added in https://github.com/Vector35/debugger/commit/2e0c5f6d031bf4a2e30c0333715fbafbfd714115

PR #941, opened 2025-12-23
Added Conditional Breakpoint Support

### LINKED ISSUE #93: Conditional breakpoints
Have the ability to add conditions to breakpoints.
What do we want for conditions? Could use either python or the expression parser.

### LINKED ISSUE #422: Support callback when a breakpoint is hit
Currently, we already support callback for the target stops. However, we should further analyze the reason of the stop and if the stop is due to a breakpoint, call a callback. There is useful in various situation, e.g., implementing conditional breakpoint.

3rdit:
## PR Overview

This PR adds conditional breakpoint support to the debugger, allowing a user to specify expressions that must evaluate to true for the breakpoint to actually trigger. This is implemented at the debugger core level (not adapter level), so it works the same across all adapters.

As discussed with @xusheng6:

The conditions for breakpoints are written using the expression parser (i.e. `$rax == 0`) making it the exact same for each debugger. The API now has:

- C++: `SetBreakpointCondition()`, `GetBreakpointCondition()` on DebuggerController
- Python: `set_breakpoint_condition()`, `get_breakpoint_condition()`

For example:
```cpp
controller->SetBreakpointCondition(address, "$rax == 0");
std::string cond = controller->GetBreakpointCondition(address);
```

```python
controller.set_breakpoint_condition(addr, "$rbx > 100")
cond = controller.get_breakpoint_condition(addr)
```

In the breakpoint widget, there's a new "Condition" column that can be set by right-clicking and pressing "Edit Condition...", this works in the widget and disassembly view.

image

The condition is also persistent across sessions (saved to the binary view metadata).

## Upcoming Changes
- ~~Condition validation on the input dialog; the idea being to stop users from entering invalid expressions~~ (this is tricky to add due to the way the expression parser works, not going to do this)
- Possibility of adding hit counts, e.g., break after N hits
- ~~It makes sense to have some sort of unit testing for this. I reviewed the unit testing and noticed how breakpoints get tested in `debugger_test.py`, it would be trivial to add a quick unit test for conditions.~~
- Adding callbacks for conditional breakpoints being hit (#422). That may be out of scope for this PR albeit and @xusheng6 will need to advise.

This PR will close #93 and maybe close #422 (if the feature above is approved to be added in this PR).

### DIFF
--- a/core/debuggerstate.cpp
+++ b/core/debuggerstate.cpp
@@ -18,7 +18,6 @@ limitations under the License.
 #include <thread>
 #include <utility>
 #include <filesystem>
-#include <optional>
 #include "lowlevelilinstruction.h"
 #include "mediumlevelilinstruction.h"
 #include "highlevelilinstruction.h"
@@ -514,8 +513,40 @@ std::vector<DebugModule> DebuggerModules::GetAllModules()
 
 
 DebuggerBreakpoints::DebuggerBreakpoints(DebuggerState* state, std::vector<ModuleNameAndOffset> initial) :
-	m_state(state), m_breakpoints(std::move(initial))
-{}
+	m_state(state)
+{
+	for (const auto& addr : initial)
+		m_breakpoints.push_back({addr, true, ""});
+}
+
+
+BreakpointEntry* DebuggerBreakpoints::FindBreakpoint(const ModuleNameAndOffset& address)
+{
+	if (m_state->GetAdapter())
+	{
+		const uint64_t targetAbsolute = m_state->GetModules()->RelativeAddressToAbsolute(address);
+		for (auto& bp : m_breakpoints)
+		{
+			if (m_state->GetModules()->RelativeAddressToAbsolute(bp.address) == targetAbsolute)
+				return &bp;
+		}
+	}
+	else
+	{
+		for (auto& bp : m_breakpoints)
+		{
+			if (bp.address == address)
+				return &bp;
+		}
+	}
+	return nullptr;
+}
+
+
+const BreakpointEntry* DebuggerBreakpoints::FindBreakpoint(const ModuleNameAndOffset& address) const
+{
+	return const_cast<DebuggerBreakpoints*>(this)->FindBreakpoint(address);
+}
 
 
 bool DebuggerBreakpoints::AddAbsolute(uint64_t remoteAddress)
@@ -524,7 +555,6 @@ bool DebuggerBreakpoints::AddAbsolute(uint64_t remoteAddress)
 		return false;
 
 	bool result = false;
-	// Always add the breakpoint as long as the adapter is connected, even if it may be already present
 	if (m_state->IsConnected())
 	{
 		m_state->GetAdapter()->AddBreakpoint(remoteAddress);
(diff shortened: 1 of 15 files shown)


xusheng6:
Hi @3rdit I will review this ASAP and let you know what I think

For context, I am working on a different PR that adds hardware breakpoint #53. And in light of the hardware breakpoint, I plan to make another large refactor on the breakpoint handling, e.g., the added breakpoints will be referred to by their index, not their address, etc. 

This would mean some non-trivial merge conflict in the future. However, I hope to minimize the frictions for you as you are doing these all for free. I will most likely review and merge your PR first, and brace for the merge conflicts on my branch. Hopefully that will be manageable

core/debuggerstate.h
```diff
@@ -84,6 +84,7 @@ namespace BinaryNinjaDebugger {
 		DebuggerState* m_state;
 		std::vector<ModuleNameAndOffset> m_breakpoints;
 		std::map<ModuleNameAndOffset, bool> m_enabledState;
+		std::map<ModuleNameAndOffset, std::string> m_conditions;
```
xusheng6: I am not a big fan of `m_conditions` here. I know you are copying the `m_enabledState` pattern, but that is really just a quick and dirty way to add "enable/disable" breakpoint into the product without a full refactor (a refactor on the breakpoint is already planned, but we wanted to slip that feature into the debugger before the refactor lands).

Ideally we would want something like this 

```
  struct Breakpoint {
      ModuleNameAndOffset address;
      bool enabled = true;
      std::string condition;
  };
```

which is apparently better in terms of how to keep things more organized. I will make a decision on whether or not to integrate the refactor with this PR

xusheng6:
@3rdit I have a look at the code today and it looks good to me in general. I also did functionality tests and it works like a charm. I will see how the unit test goes

Most likely I will take over this PR and make some changes to it, and then merge it. I will keep you posted. Thanks again for the contribution!

core/debuggerstate.h
3rdit: I like this idea. I've followed your code patterns to keep it all consistent.

I've refactored the three separate variables into a `BreakpointEntry` struct (named to avoid a conflict with the enum) stored in a single vector. `FindBreakpoint` will now return an iterator directly to avoid redundant lookups when modifying or erasing entries. This reduces complexity in the lookups/logic and should make your planned refactor easier to do.

See: https://github.com/Vector35/debugger/pull/941/commits/9eb418bdec110153f0629b13be9dab021fdd4f15 and https://github.com/Vector35/debugger/pull/941/commits/b322922244f49b45a36d56a95f47bb2ecaad16f1

core/debuggerstate.cpp
```diff
@@ -695,29 +702,83 @@ bool DebuggerBreakpoints::ContainsAbsolute(uint64_t address)
...
+
+
+std::string DebuggerBreakpoints::GetConditionAbsolute(const uint64_t address)
+{
+	for (const auto& bp : m_breakpoints)
+	{
+		if (m_state->GetModules()->RelativeAddressToAbsolute(bp.address) == address)
+			return bp.condition;
+	}
+	return "";
```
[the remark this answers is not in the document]
xusheng6: This is fine -- we can make the m_breakpoint a key value point like ModuleAndOffset -> BreakpointInfo, but since we usually have a relatively small number of breakpoints, this is not a big deal

xusheng6:
I have merged the PR, thanks a lot for the contribution! @3rdit

xusheng6:
Reopening this due to a few issues I just identified:

1. It does not work when we step onto an instruciton with conditions on it. Normally we would expect it to stop, but it will only  stop when the condition is true. This is largely because LLDB (incorrectly) reports the stop reason as Breakpoint than SingleStep. We need to find a way to workaround it
2. This may behave badly when the user is stepping on the ILs
3. We should probably ship it along side: https://github.com/Vector35/debugger/issues/942

PR #943, opened 2025-12-27
Fix conditional breakpoint evaluation when stepping onto breakpoint

3rdit:
@xusheng6 identified that LLDB will incorrectly report the stop reason as `Breakpoint` instead of `SingleStep` when stepping onto an instruction that has a breakpoint set. This caused conditional breakpoints to evaluate their condition even when the user explicitly stepped to that location which continues execution unexpectedly. https://github.com/Vector35/debugger/issues/941#issuecomment-3692548002

This fix tracks the last adapter operation and skips condition evaluation when the operation was a step (StepInto, StepOver, StepReturn, or their other variants). Condition evaluation now happens only when hitting breakpoints during Go/Continue operations.

### DIFF
--- a/core/debuggercontroller.cpp
+++ b/core/debuggercontroller.cpp
@@ -2014,7 +2014,15 @@ void DebuggerController::DebuggerMainThread()
 			AddRegisterValuesToExpressionParser();
 			AddModuleValuesToExpressionParser();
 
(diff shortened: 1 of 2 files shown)


xusheng6:
@3rdit This looks reasonable to me. Until I spend the time to actually figure out why LLDB bugs on that, this is the way to go!

PR #944, opened 2026-01-01
Add multi-row selection and copy support to registers widget

3rdit:
This PR improves the registers widget UX by adding multi-row selection and better copy/paste support.

Previously, the widget only allowed single-cell selection which made it annoying to copy, paste, modify and work with multiple registers at once. Now, a user can select multiple rows using Ctrl+Click and Shift+Click, such as any standard list.

- When selecting multiple, you can "Open Each in New Pane" which will go through each and open them as a new pane.
- I've made it so you can "Copy Row" using Ctrl+Shift+C which will copy the selected row(s) in `name: 0xvalue (hint)\n` format. 
- Setting to zero works on all selected registers instead of just one
- Select All is now a right-click feature to give that standard feature

Actions that only make sense for a single register (Edit Value, Jump to Address, Paste, etc.) are only available when working with one register, and disabled on multiple rows.

Along the way, `intx::from_string` I found doesn't like any non-numeric text and will cause Binary Ninja to crash when pasting. I've fixed this with a try-catch.

### DIFF
--- a/ui/registerswidget.cpp
+++ b/ui/registerswidget.cpp
@@ -312,7 +312,15 @@ bool DebugRegistersListModel::setData(const QModelIndex& index, const QVariant& 
 	}
 	else
 	{
-		newValue = intx::from_string<intx::uint512>(valueStr.toStdString());
+		// intx::from_string throws on invalid input (e.g., non-numeric strings)
+		try
+		{
+			newValue = intx::from_string<intx::uint512>(valueStr.toStdString());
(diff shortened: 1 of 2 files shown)


xusheng6:
@3rdit thanks for the PR! I will have a look at it ASAP (and the other one you created earlier)

xusheng6:
Thanks a lot, this is perfect! @3rdit

PR #946, opened 2026-01-05
add container file detection and extraction (and small cast issue)

### LINKED ISSUE #925: Detection For Container File With Debugger
Add support for detection of a container file and either:

- Give a hint that debugging requires the embedded file to be extracted to disk
- Or offer to write the file out to disk automatically, then debug.

3rdit:
Adds automated detection of container files (BNDB, zip archives, virtual paths) at debug launch time, and if the original file is missing for them on the disk, prompts the user to automatically extract the binary to the disk. This makes the flow a lot nicer!

When reproducing the issue I realised the `CreateProcess2 failed: 0x..` error code was wrong because they were signed `HSRESULT` values, now they are casted to `uint32_t`.

This PR will close #925

### DIFF
--- a/ui/controlswidget.cpp
+++ b/ui/controlswidget.cpp
@@ -28,6 +28,9 @@ limitations under the License.
 #include <thread>
 #include "progresstask.h"
 #include "attachprocess.h"
+#include <filesystem>
+#include <fstream>
+#include <QFileDialog>
 
 using namespace BinaryNinjaDebuggerAPI;
 using namespace BinaryNinja;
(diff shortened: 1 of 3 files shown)


ui/controlswidget.cpp
```diff
@@ -168,8 +171,86 @@ QString DebugControlsWidget::getToolTip(const QString& name)
...
+		.arg(QString::fromStdString(originalPath));
+
+	if (QMessageBox::question(this, "File Not Found", prompt) != QMessageBox::Yes)
+		return false;
+
+	fs::path defaultPath = fs::current_path() / fs::path(originalPath).filename();
+	if (isBndb)
+		defaultPath = fs::path(currentPath).parent_path() / fs::path(originalPath).filename();
+
+	QString extractPath = QFileDialog::getSaveFileName(
```
xusheng6, line 235: This would prompt the dialog and default to the binaryninja is running from. It would be better to set the default directory to be the path the zip file is in: https://docs.binary.ninja/dev/containertransforms.html#metadata-and-virtual-paths

ui/controlswidget.cpp
```diff
@@ -168,8 +171,86 @@ QString DebugControlsWidget::getToolTip(const QString& name)
...
+	std::string originalPath = file->GetOriginalFilename();
+
+	bool isBndb = currentPath.ends_with(".bndb") && currentPath != originalPath;
+	bool isVirtualPath = currentPath.find("::") != std::string::npos;
+	bool isVirtualFile = !fs::path(originalPath).is_absolute();
+
+	if (!isBndb && !isVirtualPath && !isVirtualFile)
+		return true;
+
+	if (fs::exists(originalPath))
```
xusheng6: It should check if the current executable path exists instead of that from the originalPath. The user could have already extracted it somewhere and updated the executable path, and this check would miss it

ui/controlswidget.cpp
```diff
@@ -168,8 +171,86 @@ QString DebugControlsWidget::getToolTip(const QString& name)
...
+	auto file = data->GetFile();
+
+	std::string execPath = m_controller->GetExecutablePath();
+	if (!execPath.empty() && fs::exists(execPath))
+		return true;
+
+	std::string currentPath = file->GetFilename();
+	std::string originalPath = file->GetOriginalFilename();
+
+	bool isBndb = currentPath.ends_with(".bndb") && currentPath != originalPath;
```
xusheng6: Use file->IsBackedByDatabase() instead of checking the file extension

xusheng6 (changes requested):
This looks great! I tested and it works smoothly. I have made a few comments -- once those are addressed this PR is good to go!

ui/controlswidget.cpp
3rdit: Fixed in https://github.com/Vector35/debugger/pull/946/commits/739b802f2529a17d2535896bb3c8a393a50f2033

ui/controlswidget.cpp
3rdit: Fixed in https://github.com/Vector35/debugger/pull/946/commits/739b802f2529a17d2535896bb3c8a393a50f2033

ui/controlswidget.cpp
3rdit: Fixed in https://github.com/Vector35/debugger/pull/946/commits/739b802f2529a17d2535896bb3c8a393a50f2033

ui/controlswidget.cpp
xusheng6: I reverted the relevant changes. They are not robust enough. Maybe we should offer sth like a file_on_disk for this purpose

PR #947, opened 2026-01-06
Add nullptr checks for dynamically acquired UI objects

### LINKED ISSUE #722: Fix missing nullptr checks
The debugger codebase is known to have many missing nullptr checks. Although I originally believed that it is not worthwhile to try to fix them since they should not really be null at all (or it may indicate another big or issue), I recently ran into a case where a subtle difference in the interaction could cause a nullptr (and then a crash), which I did not think of: https://github.com/Vector35/debugger/commit/9334e78afc9ae683a15751c866ded5df80312407

The reason why we should really add this nullptr checks is that the debugger is dynamic in nature, and it is hard to always be smart to think of all of the possible interactions

3rdit:
This PR adds nullptr checks for dynamically acquired objects `UIContext`, `ViewFrame`, `View` and `BinaryView`.  This will prevent crashes when UI state changes during something like closing a tab as an event happens.

This PR closes #722

### DIFF
--- a/ui/breakpointswidget.cpp
+++ b/ui/breakpointswidget.cpp
@@ -454,7 +454,13 @@ void DebugBreakpointsWidget::jump()
 
 	BreakpointItem bp = m_model->getRow(sel[0].row());
 	UIContext* context = UIContext::contextForWidget(this);
+	if (!context)
+		return;
+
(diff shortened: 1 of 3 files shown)


xusheng6:
Thanks @3rdit!

PR #950, opened 2026-01-07
Add hardware breakpoint support

xusheng6:
Fix github.com/Vector35/debugger/isses/53

### DIFF
--- a/api/python/debuggercontroller.py
+++ b/api/python/debuggercontroller.py
@@ -371,20 +371,24 @@ class DebugBreakpoint:
     * ``address``: the absolute address of the breakpoint
     * ``enabled``: whether the breakpoint is enabled (read-only)
     * ``condition``: the condition expression for the breakpoint (empty if no condition)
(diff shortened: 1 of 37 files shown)


api/python/debuggercontroller.py
```diff
@@ -371,28 +371,33 @@ class DebugBreakpoint:
     * ``address``: the absolute address of the breakpoint
     * ``enabled``: whether the breakpoint is enabled (read-only)
     * ``condition``: the condition expression for the breakpoint (empty if no condition)
+    * ``type``: the type of breakpoint (Software, HardwareExecute, HardwareRead, HardwareWrite, HardwareAccess)
+    * ``size``: the size in bytes for hardware breakpoints/watchpoints (1, 2, 4, or 8)
 
     """
-    def __init__(self, module, offset, address, enabled, condition=""):
+    def __init__(self, module, offset, address, enabled, condition="", bp_type=DebugBreakpointType.BNSoftwareBreakpoint, size=1):
```
plafosse: Seems like this should really be a dataclass. I think this would clean up this class a bunch.

PR #969, opened 2026-01-21
Manual rebasing setting and rebase to remote base action

### LINKED ISSUE #214: Support not rebasing the input binary view
For a PIE executable, executing under the debugger causes all addresses to permanently change to their rebased runtime values.

### LINKED ISSUE #821: Support manually rebasing the input binary view to match the debugger status
The debugger tries to automatically find out the remote base and rebase the binary accordingly. However, when it fails, there is no way for the user to manually rebase it

3rdit:
This PR adds the ability to disable automatic rebasing when debugging and allow (optional) manual rebasing instead.

A new setting `debugger.autoRebase` (default: true) controls whether the debugger automatically rebases the input binary view when the module is loaded. The user can manually rebase via "Rebase to Remote Base" in the debugger menu or context menu.

This PR will close #821, will close #214

### DIFF
--- a/api/debuggerapi.h
+++ b/api/debuggerapi.h
@@ -769,6 +769,9 @@ namespace BinaryNinjaDebuggerAPI {
 		uint64_t RelativeAddressToAbsolute(const ModuleNameAndOffset& address);
 		ModuleNameAndOffset AbsoluteAddressToRelative(uint64_t address);
 
+		// rebasing
+		bool RebaseToRemoteBase();
+
 		size_t RegisterEventCallback(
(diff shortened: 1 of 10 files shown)


xusheng6:
Hi @3rdit , thanks for the PR! I had a quick look at the PR, and it looks great. However, I prefer that during manual rebasing, a base address can be provided by the user. An important motivation for this PR is that when the auto-detected base address is wrong, the user would still be able to provide the correct value after inspecting the target. Bonus point if we provide a default value that is likely the one the debugger thinks is the right base, but that is totally optional

xusheng6:
Thanks for the PR @3rdit . I merged it with some minor mods

PR #1007, opened 2026-03-02
Implement GetTTDCallsForSymbols for esReven adapter

razaina:
Sends rvn:get-calls-by-symbol:<symbol>[:<start_ret>:<end_ret>] packet
and parses the JSON response into TTDCallEvent objects, enabling
Binary Ninja's TTD Calls widget to query call history from REVEN.

### DIFF
--- a/core/adapters/esrevenadapter.cpp
+++ b/core/adapters/esrevenadapter.cpp
@@ -1990,6 +1990,125 @@ Ref<Settings> EsrevenAdapterType::RegisterAdapterSettings()
 }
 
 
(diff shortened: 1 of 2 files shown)


xusheng6:
So the implementation of the function `EsrevenAdapter::GetTTDCallsForSymbols` slipped in via an earlier commit 0e0dc2a38ee530533ce2ee7ec5669143fa108c8b, so I cherry-picked the difference into cfb69b4e590b22d7631fab3723c0039185addc05. Will close this PR

Issue #1021, opened 2026-03-19
GDB MI adapter does not parse XMM register values properly


xusheng6:
```
[Default] Failed to parse GDB value: "<unavailable>"
[Default] Failed to parse GDB value: "<unavailable>"
[Default] Failed to parse GDB value: "{v4_float = {0x00000000, 0x00000000, 0x00000000, 0x00000000}, v2_double = {0x0000000000000000, 0x0000000000000000}, v16_int8 = {0x00 <repeats 16 times>}, v8_int16 = {0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000, 0x0000}, v4_int32 = {0x00000000, 0x00000000, 0x00000000, 0x00000000}, v2_int64 = {0x0000000000000000, 0x0000000000000000}, uint128 = 0x00000000000000000000000000000000}"
[Default] Failed to parse GDB value: "{v4_float = {0x6d6f682f, 0x75782f65, 0x6e656873, 0x65642f67}, v2_double = {0x75782f656d6f682f, 0x65642f676e656873}, v16_int8 = {0x2f, 0x68, 0x6f, 0x6d, 0x65, 0x2f, 0x78, 0x75, 0x73, 0x68, 0x65, 0x6e, 0x67, 0x2f, 0x64, 0x65}, v8_int16 = {0x682f, 0x6d6f, 0x2f65, 0x7578, 0x6873, 0x6e65, 0x2f67, 0x6564}, v4_int32 = {0x6d6f682f, 0x75782f65, 0x6e656873, 0x65642f67}, v2_int64 = {0x75782f656d6f682f, 0x65642f676e656873}, uint128 = 0x65642f676e65687375782f656d6f682f}"
[...]

PR #1022, opened 2026-03-19
Fix GDB MI register parsing for XMM/YMM and unavailable values

xusheng6:
## Summary
- Handle `<unavailable>` register values gracefully (return 0 instead of logging an error)
- Parse XMM composite struct values by extracting the `uint128 = 0x...` field
- Parse YMM composite struct values by combining `v2_int128 = {0xHIGH, 0xLOW}` fields into a 256-bit value

## Test plan
- [ ] Attach to a gdbserver target and verify no "Failed to parse GDB value" errors for vector registers
- [ ] Verify XMM register values are correctly displayed
- [ ] Verify YMM register values are correctly displayed
- [ ] Test with a target that has unavailable registers


xusheng6:
Fixes: https://github.com/Vector35/debugger/issues/1021

PR #1026, opened 2026-03-20
Fixed general bugs

3rdit:
Took some time to explore the codebase a bit and cleaned up a few issues I found.

1. `detach()` called `BNDebuggerQuit` instead of `BNDebuggerDetach`
2. `pid_attach` setter was broken and didn't use the correct decorator (was a copy-paste of the port setter stuff), likewise with the doc for it.
3. `get_addr_info` wasn't usable due to the fact a plain int was passed where the FFI expected a 64-byte buffer
4. `__hash__` used (pid, pid) instead of (pid, name) which never matched the `__eq__`
5. `SuspendThread`/`ResumeThread` never persisted freeze state, `GetAllThreads()` returns by value, so `m_isFrozen` was set on a local copy
6. `DetachAndWait`/`QuitAndWait` leaked mutex on early return due to `m_targetControlMutex` staying locked if `IsConnected()` returned false
7. `RunTo` sent duplicated stop events to the UI, `NotifyStopped` is called in both *Internal and *AndWait, unlike any other operation
8. `DbgEngTTDAdapter::Start()` didn't `return false`, so may cause it to fall through - this is now fixed.
10. In GDB, `substr(4)` would have corrupted the `monitor` command, now it's `substr(8)`.
11. Initialise `m_parentItem` pointer to avoid possible garbage read (unlikely but also just consistency)

### DIFF
--- a/api/python/debuggercontroller.py
+++ b/api/python/debuggercontroller.py
@@ -83,7 +83,7 @@ class DebugProcess:
         return not (self == other)
 
     def __hash__(self):
-        return hash((self.pid, self.pid))
+        return hash((self.pid, self.name))
 
     def __setattr__(self, name, value):
         try:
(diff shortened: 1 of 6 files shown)


xusheng6:
Thanks a lot! @3rdit

PR #1045, opened 2026-04-06
Fix crash from joining the event thread on itself

### LINKED ISSUE #1038: crashpad::`anonymous namespace'::HandleAbortSignal
Sentry Issue: [BINARYNINJA-2K](https://vector35.sentry.io/issues/7377973311/?referrer=github_integration)

```
STATUS_FATAL_APP_EXIT / 0x7ff6e3e558c6: Fatal Error: STATUS_FATAL_APP_EXIT / 0x7ff6e3e558c6
  File "crashpad_client_win.cc", line 230, in crashpad::`anonymous namespace'::HandleAbortSignal
  File "throw.cpp", line 79, in _CxxThrowException
  File "thread0.cpp", line 33, in std::_Throw_Cpp_error
  File "thread", line 127, in std::thread::join
  File "debuggercontroller.cpp", line 50, in BinaryNinjaDebugger::DebuggerController::~DebuggerController
...
(22 additional frame(s) were not displayed)
```

@bdash says:
> The immediate cause of the crash is that `DebuggerController`'s destructor calls `std::thread::join` on `m_debuggerEventThread` while it is executing on the debugger event thread. Attempting to join the current thread results in `std::system_error(std::errc::resource_deadlock_would_occur)` being thrown, and since that exception is not handled the app exits.
>
> `DebuggerController` is destroyed on this thread because the `DebuggerFileAccessor` is the final reference to the controller, and [it is deleted](https://github.com/Vector35/debugger/blob/f7c3174499bbdd38558ba136641c30169606ebff/core/debuggercontroller.cpp#L1878) in response to [`TargetExitedEventType`, `DetachedEventType`, and `LaunchFailureEventType`](https://github.com/Vector35/debugger/blob/f7c3174499bbdd38558ba136641c30169606ebff/core/debuggercontroller.cpp#L1859-L1862)
>

xusheng6:
## Summary
- Defer `delete m_accessor` in `EventHandler` to a worker thread via `BinaryNinja::WorkerEnqueue`
- The `DebuggerFileAccessor` holds a `DbgRef<DebuggerController>`, and when it's the last reference, deleting it on the event thread triggers `~DebuggerController` which calls `m_debuggerEventThread.join()` — deadlocking because we ARE the event thread
- By deferring the deletion, the controller's destructor runs on a worker thread where `join()` succeeds normally

## Test plan
- [ ] Debug a target, let it exit normally — verify no crash on exit
- [ ] Debug a target, detach — verify no crash on detach
- [ ] Trigger a launch failure (e.g., invalid remote address) — verify no crash

Fixes #1038

### DIFF
--- a/core/debuggercontroller.cpp
+++ b/core/debuggercontroller.cpp
@@ -1875,13 +1875,18 @@ void DebuggerController::EventHandler(const DebuggerEvent& event)
 
 		if (m_accessor)
 		{
-			// Defer deletion to a worker thread. The accessor holds a DbgRef<DebuggerController>,
+			// Defer deletion to a detached thread. The accessor holds a DbgRef<DebuggerController>,
 			// and if it is the last reference, deleting it here (on the event thread) would trigger
-			// ~DebuggerController which calls m_debuggerEventThread.join() -- deadlocking because
-			// we ARE the event thread.
+			// ~DebuggerController which calls m_debuggerEventThread.join() -- deadlocking/crashing
+			// because we ARE the event thread.
+			//
+			// This can happen when Destroy() races with event processing: EventHandler sets
+			// ConnectionStatus to NotConnected (line above), and another thread observes this,
+			// calls Destroy() which removes the global array ref, making the accessor's DbgRef
+			// the last reference to the controller.
 			auto* accessor = m_accessor;
 			m_accessor = nullptr;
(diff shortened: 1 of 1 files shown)


xusheng6:
Recreating with test_ branch prefix for CI.

Issue #1032, opened 2026-03-31
Debugger should not focus the debugger sidebar every time the target executes


xusheng6:
This makes it really awkward to use the debugger in conjunction with sidekick

PR #1048, opened 2026-04-06
Fix use-after-free crash in UI callbacks on rebase

### LINKED ISSUE #1039: BinaryNinjaDebugger::DebuggerUICallbacks::NotifyRebaseBinaryView
Sentry Issue: [BINARYNINJA-2H](https://vector35.sentry.io/issues/7361879204/)

```
SIGSEGV / SI_KERNEL / 0x0: Fatal Error: SIGSEGV / SI_KERNEL / 0x0
  File "debuggercontroller.cpp", line 4514, in BinaryNinjaDebugger::DebuggerUICallbacks::NotifyRebaseBinaryView
  File "debuggercontroller.cpp", line 4579, in BinaryNinjaDebugger::DebuggerController::RebaseToAddress
```

@bdash says:
> I think this is due to a lifetime issue with the debugger's UI callbacks. `DebuggerUI::DebuggerUI` calls `DebuggerController::SetDebuggerUICallbacks` and passes a pointer stored in `m_uiCallbacks`. `DebuggerUI::~DebuggerUI` deletes that object, but nothing unregisters the callbacks from the `DebuggerController`. I'm not sure what, if anything prevents the `DebuggerController` from making a UI callback after the `DebuggerUI` goes away.

xusheng6:
## Summary
- Unregister UI callbacks from the `DebuggerController` in `~DebuggerUI()` before deleting them, preventing a use-after-free crash when the controller processes a rebase after the UI is destroyed.
- `SetDebuggerUICallbacks` now accepts `nullptr` to clear the callbacks. The existing null check in `RebaseToAddress()` ensures any rebase attempt after UI destruction safely returns `false`.

## Test plan
- [ ] Launch a debug target, then close the binary view tab before the launch completes. Verify no crash occurs when the target eventually stops and triggers a rebase.
- [ ] Normal debugging workflow (launch, hit breakpoints, rebase) still works correctly with UI open.

Fixes #1039

### DIFF
--- a/ui/ui.cpp
+++ b/ui/ui.cpp
@@ -1709,6 +1709,9 @@ DebuggerUI::DebuggerUI(UIContext* context, DebuggerControllerRef controller) :
 
 DebuggerUI::~DebuggerUI()
 {
+	// Unregister UI callbacks from the controller before deleting them, so the controller
+	// does not invoke a dangling callback if events arrive after the UI is destroyed.
+	m_controller->SetDebuggerUICallbacks(nullptr);
 	if (m_uiCallbacks)
 	{
 		delete m_uiCallbacks;
--- a/api/debuggercontroller.cpp
+++ b/api/debuggercontroller.cpp
@@ -1001,7 +1001,10 @@ void DebuggerController::RemoveEventCallback(size_t index)
 
 void DebuggerController::SetDebuggerUICallbacks(DebuggerUICallbacks* cb)
 {
-	BNDebuggerSetDebuggerUICallbacks(m_object, cb->GetCallbacks(), cb);
+	if (cb)
+		BNDebuggerSetDebuggerUICallbacks(m_object, cb->GetCallbacks(), cb);
+	else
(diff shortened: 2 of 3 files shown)


ui/ui.cpp
```diff
@@ -1828,7 +1831,6 @@ void DebuggerUI::navigateDebugger(uint64_t address)
 			frame->navigate(m_controller->GetData(), address, true, true);
 	}
 
-	openDebuggerSideBar(frame);
```
plafosse: Whats the purpose of this? Seems like a behavior change.

xusheng6: This is a change introduced earlier: https://github.com/Vector35/debugger/issues/1032. The reason that I wanted it to no longer focus the debugger sidebar is that if the user is using sidekick to automate the debugger, this will cause the sidekick sidebar to lose focus. 

I think I made some git mistake so that it gets wiped out unexpectedly. I will add it back

xusheng6: I force-pushed to rebase it on top of dev. dev now already has the change

plafosse (approved):
This one looks good now.

PR #1062, opened 2026-04-10
Fix AttachProcessDialog leak causing DebuggerController ref leak

### LINKED ISSUE #1061: AttachProcessDialog leaks when the user opens it from the menu, not the controls widget
## Summary

`AttachProcessDialog` is allocated on the heap but never freed in `ui/ui.cpp`. The dialog owns a `ProcessListWidget` which holds a `DbgRef<DebuggerController>`, causing the controller to also leak.

## Root Cause

In `ui/ui.cpp` (around line 858), the attach action creates the dialog with `new` but never deletes it:

```cpp
auto dialog = new AttachProcessDialog(context->mainWindow(), controller);
if (dialog->exec() != QDialog::Accepted)
    return;

uint32_t pid = dialog->GetSelectedPid();
```

Unlike `AdapterSettingsDialog` which sets `Qt::WA_DeleteOnClose` in its constructor, `AttachProcessDialog` has no cleanup mechanism.

## Stack trace from leak detector

```
Possible leak of BinaryNinjaDebuggerAPI::DebuggerController, referenced at:
0   libdebuggerui.dylib  ProcessListWidget::ProcessListWidget(QWidget*, ...) + 164
1   libdebuggerui.dylib  AttachProcessDialog::AttachProcessDialog(QWidget*, ...) + 312
2   libdebuggerui.dylib  GlobalDebuggerUI::SetupMenu(UIContext*)::$_23 + 636
```

## Fix

Add `dialog->deleteLater()` after the last use of the dialog to ensure proper cleanup.

## Related

Identified by @bdash in https://github.com/Vector35/debugger/issues/1042#issuecomment-4216618399

xusheng6:
## Summary
- `AttachProcessDialog` was heap-allocated with `new` but never freed in the attach action handler (`ui/ui.cpp`)
- Since it owns a `ProcessListWidget` which holds a `DbgRef<DebuggerController>`, this also caused the controller to leak
- Added `deleteLater()` on all exit paths after the dialog is used

## Test plan
- [ ] Open a binary, attach to a process via the debugger, then stop debugging and close the file
- [ ] Verify no `DebuggerController` leak is reported in the leak detector

Fix #1061

### DIFF
--- a/ui/ui.cpp
+++ b/ui/ui.cpp
@@ -857,9 +857,13 @@ void GlobalDebuggerUI::SetupMenu(UIContext* context)
 
 				auto dialog = new AttachProcessDialog(context->mainWindow(), controller);
 				if (dialog->exec() != QDialog::Accepted)
+				{
+					dialog->deleteLater();
 					return;
+				}
 
 				uint32_t pid = dialog->GetSelectedPid();
+				dialog->deleteLater();
 				if (pid == 0)
 					return;
 
--- a/ui/ui.cpp
+++ b/ui/ui.cpp
@@ -855,15 +855,11 @@ void GlobalDebuggerUI::SetupMenu(UIContext* context)
 						return;
 				}
 
-				auto dialog = new AttachProcessDialog(context->mainWindow(), controller);
-				if (dialog->exec() != QDialog::Accepted)
-				{
-					dialog->deleteLater();
+				AttachProcessDialog dialog(context->mainWindow(), controller);
+				if (dialog.exec() != QDialog::Accepted)
 					return;
-				}
 
-				uint32_t pid = dialog->GetSelectedPid();
-				dialog->deleteLater();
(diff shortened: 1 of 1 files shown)


plafosse:
Why allocate at all?

```
AttachProcessDialog dialog(context->mainWindow(), controller);
if (dialog.exec() != QDialog::Accepted)
    return;

uint32_t pid = dialog.GetSelectedPid();
if (pid == 0)
    return;
```

xusheng6:
> Why allocate at all?
> 
> 
> 
> ```
> 
> AttachProcessDialog dialog(context->mainWindow(), controller);
> 
> if (dialog.exec() != QDialog::Accepted)
> 
>     return;
> 
> 
> 
> uint32_t pid = dialog.GetSelectedPid();
> 
> if (pid == 0)
> 
>     return;
> 
> ```

Good point -- let me fix that

PR #1064, opened 2026-04-14
added the use settings next time checkbox to adapter settings dialog

### LINKED ISSUE #973: Debugger adapter settings dialog should only be skipped if "use same settings next time" checkbox is checked
Right now it works like: if the user launch/attach/connect for the first time, we show the dialog, and in subsequent ones, we skip it entirely. This can be surprising to the users. We should add a "Use same settings next time" checkbox into the dialog, and only if it is checked, we skip the dialog the next time

3rdit:
Adds a "Use same settings next time" checkbox to the adapter settings (enabled on default) dialog that will skip the dialog silently if checked.

Closes #973

### DIFF
--- a/api/debuggerapi.h
+++ b/api/debuggerapi.h
@@ -830,6 +830,16 @@ namespace BinaryNinjaDebuggerAPI {
 		bool IsFirstConnectToDebugServer();
 		bool IsFirstAttach();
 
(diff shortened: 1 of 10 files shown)


xusheng6:
Closing in favor of #1133, thanks for the contribution!

We will take it forward from there and your contribution will be preserved when we merge it

Issue #1089, opened 2026-05-20
Simplify adapter-vs-controller stop plumbing after worker-queue refactor


xusheng6:
Follow-up to #1088 / #1087.

## What is the adapter-vs-controller stop distinction?

The controller currently distinguishes two kinds of stop:

- **Adapter stop** — the underlying engine (DbgEng, LLDB, gdbstub, …) reports that the target has stopped. Posted as `AdapterStoppedEventType` events.
- **Controller stop** — the moment the debugger hands control back to the user. Posted as `TargetStoppedEventType` events via `NotifyStopped`.

The two can diverge because:

1. **Conditional breakpoints** — adapter says "stopped at a breakpoint," controller evaluates the condition, and if it's false, silently resumes (`debuggercontroller.cpp:2218-2256`). The user never sees the adapter stop.
2. **IL stepping** — a single user "Step Into HLIL" causes N underlying single-step adapter stops. Only the final one, where the IP lands on the right IL boundary, should be user-visible.
3. **External adapter commands** — when the user types e.g. `si` directly into the LLDB REPL, no controller-driven op is in flight to consume the adapter stop. The dispatcher synthesizes a `TargetStoppedEvent` for the UI (`debuggercontroller.cpp:2276-2297`).

The semantic need is genuine — these features can't go away.

## What's awkward about the current implementation

The mechanism that implements this distinction is more complicated than the semantics require:

- `AdapterStoppedEventType` is a public event type going through the same `PostDebuggerEvent` queue as user-visible events.
- `m_lastAdapterStopEventConsumed` is a flag the dispatcher and `ExecuteAdapterAndWait` pass back and forth to decide whether a stop should be "silenced" or surfaced.
- The conditional-breakpoint check runs on the **dispatcher thread**, and from there calls `m_adapter->Go()` directly. The code apologizes for this with a comment: `// Using m_adapter->Go() directly instead of Go() to avoid mutex deadlock`. The `m_suppressResumeEvent` flag exists to paper over the fact that the dispatcher is now driving the adapter.
- `ExecuteAdapterAndWait` registers a temporary event callback, blocks on a semaphore, waits for the dispatcher to fire it, then unregisters (`debuggercontroller.cpp:2860-2883`).

All of that exists because in the original design, the adapter's event thread, the controller's dispatcher thread, and a per-op spawned thread are three separate threads that need a common rendezvous — and `PostDebuggerEvent` is what they have.

## What the worker-queue refactor enables

[...]

Issue #1088, opened 2026-05-20
Refactor: route DebuggerController operations through a per-controller worker queue


xusheng6:
## Motivation

The 17 public operations on `DebuggerController` (`Launch`, `Attach`, `Connect`, `Go`, `GoReverse`, `Step{Into,Over,Return}{,Reverse}`, `RunTo{,Reverse}`, `Restart`, `Detach`, `Quit`, `Pause`) each spawn a fresh `std::thread` and detach it. Every site captures `this` raw (or via `DbgRef` after #1086 lands), then calls back into the controller. This has caused a recurring family of crashes — the controller can be destroyed while one of these detached threads is still running, leading to use-after-free on its members. Several recent Sentry reports trace to exactly this pattern:

- BINARYNINJA-3X — `RspConnector::SendRaw` via `CorelliumAdapter::BreakInto` (26 events)
- BINARYNINJA-7M — `DbgEngTTDAdapter::Reset` → `EngineLoop`
- BINARYNINJA-60 / BINARYNINJA-5Z — `DbgEngAdapter::ReadMemory`
- BINARYNINJA-1A — `DbgEngAdapter::Attach` worker thread
- #1083 — `DetachAndWait` crash

The per-call `DbgRef` capture (#1086) is a stopgap that keeps the controller alive for the duration of the detached call. The structural fix is to stop spawning per-op threads in the first place.

## Approach

Replace the per-operation detach pattern with:

1. **One persistent worker thread per controller**, owned by the controller and joined in the destructor (after draining).
2. **A FIFO work queue** that every public operation submits to. The queue serializes ops naturally — no concurrent adapter calls, no mutex dance to prevent them.
3. **`Submit()` / `SubmitAndWait()` templates** that handle queueing and re-entrant calls (when the worker invokes another operation on itself, run inline instead of deadlocking).
4. **`Pause()` and `PauseAndWait()` bypass the queue** and call `m_adapter->BreakInto()` directly on the caller's thread — the worker is blocked inside `ExecuteAdapterAndWait` for the in-flight op and needs an out-of-band signal to unblock.
5. **Timeout parameter** on every `XxxAndWait`, default `std::chrono::milliseconds::max()` so existing callers don't change behavior. Closes the long-standing \"sync ops can hang forever\" complaint without an API break.

The public API surface is unchanged — every existing `Launch()` / `LaunchAndWait()` call site keeps working. Internally:

[...]

PR #1090, opened 2026-05-20
Route DebuggerController operations through a per-controller worker queue

xusheng6:
Replaces the "spawn one detached `std::thread` per operation" model in `DebuggerController` with a single per-controller worker thread + FIFO queue, and makes the controller own its state directly instead of bouncing it through the event dispatcher.

Implements #1088 and #1089. Supersedes the auto-closed #1087.

## Why

The per-op detach pattern (17 sites) captured `this` and detached, so a controller could be destroyed while a worker was still running — the recurring use-after-free family in Sentry (#1083, BINARYNINJA-1A/3X/5Z/60/7M). The queue makes lifetime structural: the worker is owned by the controller and joined in the destructor, so no adapter work can outlive it.

## What changed

- **Worker queue.** All ops route through `Submit` / `SubmitAndWait`. One worker thread per controller, joined on destruction. `Submit` runs inline if called from the worker itself (so `Restart` → `Quit` + `Launch` doesn't self-deadlock).
- **Timeout.** Every `XxxAndWait` gains an optional `std::chrono::milliseconds timeout` (default = wait forever). Existing callers unchanged.
- **Pause / Restart / Quit / Detach** interrupt a running target out-of-band: they call `BreakInto()` on the caller's thread, then queue the actual op. Without this they'd sit behind the in-flight resume forever.
- **Adapter-stop channel.** `AdapterStoppedEvent` is now an internal worker signal (condvar), not a public dispatcher event. Conditional-breakpoint silent-resume moved onto the worker; the dispatcher's stop machinery, `m_lastAdapterStopEventConsumed`, and `m_suppressResumeEvent` are gone. Adapters are unchanged.
- **State on the worker.** `EventHandler` (the controller subscribing to its own broadcast) is replaced by `ApplyOwnStateForEvent`, called inline in `PostDebuggerEvent` before the event is broadcast. The dispatcher is now pure fan-out to external consumers (UI/plugins). This removes the worker-vs-dispatcher race that broke Restart.
- **Mutexes removed.** `m_adapterMutex` / `m_adapterMutex2` guarded concurrent adapter access that can no longer happen; replaced with an `assert(on worker thread)`.
- **Stop is always reported.** The worker op calls `NotifyStopped` even on a user-requested break (the old Pause path used to do it; the new out-of-band Pause doesn't), fixing "target halts but UI still shows running".

## API compatibility

Public method signatures unchanged (timeout is an appended optional arg). Callers in `core/ffi.cpp` and `ui/uinotification.cpp` need no changes.

## Out of scope (follow-ups)

- Adapter-internal threads (EngineLoop, LLDB listener, RSP loops) — need adapter dtors as join barriers.
- `m_targetControlMutex` is now largely redundant; kept to limit churn.
- Refcounted `DebugAdapter` was tried and reverted — the adapter has no independent lifecycle (freed only in `~DebuggerState`), so it solved no real UAF.
- UI repaint throttling for fast stepping — #1093.

## Test plan

- [x] Builds clean (`ninja` — core + ui)
- [ ] Launch / step / breakpoint / pause while running / detach mid-step / close tab mid-op
- [ ] Restart on a running target
- [ ] `si` typed directly in the LLDB REPL while idle (spontaneous stop)
- [ ] Conditional breakpoint with a false condition (silent resume)
- [ ] IL step surfaces one stop per user action
- [ ] `test/debugger_test.py`

### DIFF
--- a/core/debuggercontroller.cpp
+++ b/core/debuggercontroller.cpp
@@ -17,12 +17,17 @@ limitations under the License.
 #include "debuggercontroller.h"
 #include <thread>
 #include <fstream>
+#include "base/assertions.h"
 #include "lowlevelilinstruction.h"
 #include "mediumlevelilinstruction.h"
 #include "highlevelilinstruction.h"
 #include "debuggerfileaccessor.h"
 
 using namespace BinaryNinjaDebugger;
+
+namespace BinaryNinjaDebugger {
+	thread_local DebuggerController* t_controllerOnWorker = nullptr;
+}
 
 DebuggerController::DebuggerController(BinaryViewRef data): BinaryDataNotification(Rebased)
 {
@@ -36,14 +41,45 @@ DebuggerController::DebuggerController(BinaryViewRef data): BinaryDataNotificati
 	m_state = new DebuggerState(data, this);
 	m_adapter = nullptr;
 	m_shouldAnnotateStackVariable = Settings::Instance()->Get<bool>("debugger.stackVariableAnnotations");
-	RegisterEventCallback([this](const DebuggerEvent& event) { EventHandler(event); }, "Debugger Core");
 
 	m_debuggerEventThread = std::thread([&]{ DebuggerMainThread(); });
+
+	m_workerShouldExit = false;
+	m_workerThread = std::thread([this] { WorkerThreadMain(); });
+
+	m_interruptThread = std::thread([this] { InterruptThreadMain(); });
 }
 
 
 DebuggerController::~DebuggerController()
 {
+	// The worker can be blocked on either condition variable -- m_workQueueCv (idle in
+	// the outer loop) or m_adapterStopCv (inside WaitForAdapterStop during an op). Each
+	// CV's wait predicate reads m_workerShouldExit, so the flag must be modified while
+	// holding the matching mutex to publish the change correctly to that waiter. We hold
+	// BOTH mutexes when setting the flag (scoped_lock is deadlock-safe) and then notify
+	// both CVs. Otherwise a waiter on the CV whose mutex we didn't hold can miss the
+	// wakeup and never observe the flag, hanging this destructor on join.
+	{
+		std::scoped_lock shutdownLock(m_workQueueMutex, m_adapterStopMutex);
+		m_workerShouldExit = true;
+	}
+	m_workQueueCv.notify_all();
+	m_adapterStopCv.notify_all();
+	if (m_workerThread.joinable())
+		m_workerThread.join();
+
+	// Stop the interrupt thread before the adapter/state are torn down below: it touches
+	// m_adapter and m_state->AdapterAccessMutex() in BreakInto, both of which outlive this
+	// join but not the delete m_state further down.
+	{
+		std::lock_guard<std::mutex> interruptLock(m_interruptMutex);
+		m_interruptShouldExit = true;
+	}
+	m_interruptCv.notify_all();
+	if (m_interruptThread.joinable())
+		m_interruptThread.join();
+
 	m_shouldExit = true;
 	m_cv.notify_all();
 	if (m_debuggerEventThread.joinable())
@@ -57,6 +93,28 @@ DebuggerController::~DebuggerController()
 		delete m_state;
 		m_state = nullptr;
 	}
+}
+
+
+void DebuggerController::WorkerThreadMain()
+{
+	t_controllerOnWorker = this;
+	while (true)
+	{
+		std::function<void()> task;
+		{
+			std::unique_lock<std::mutex> lock(m_workQueueMutex);
+			m_workQueueCv.wait(lock, [this] {
+				return m_workerShouldExit || !m_workQueue.empty();
+			});
+			if (m_workerShouldExit && m_workQueue.empty())
+				break;
+			task = std::move(m_workQueue.front());
+			m_workQueue.pop();
+		}
+		task();
+	}
+	t_controllerOnWorker = nullptr;
 }
 
 
@@ -290,15 +348,13 @@ bool DebuggerController::Launch()
 	if (!CanStartDebgging())
 		return false;
 
-	DbgRef<DebuggerController> self = this;
-	std::thread([self]() { self->LaunchAndWait(); }).detach();
+	Submit([this] { LaunchAndWaitOnWorker(); });
 	return true;
 }
 
 
 DebugStopReason DebuggerController::LaunchAndWaitInternal()
 {
-	m_userRequestedBreak = false;
 
 	if (Settings::Instance()->Get<bool>("debugger.safeMode"))
 	{
@@ -330,21 +386,23 @@ DebugStopReason DebuggerController::LaunchAndWaitInternal()
 }
 
 
-DebugStopReason DebuggerController::LaunchAndWait()
+DebugStopReason DebuggerController::LaunchAndWaitOnWorker()
 {
 	// This is an API function of the debugger. We only do these checks at the API level.
 	if (!CanStartDebgging())
 		return InvalidStatusOrOperation;
 
-	if (!m_targetControlMutex.try_lock())
-		return InternalError;
-
 	auto reason = LaunchAndWaitInternal();
-	if (!m_userRequestedBreak && (reason != ProcessExited) && (reason != InternalError))
+	if ((reason != ProcessExited) && (reason != InternalError))
 		NotifyStopped(reason);
 
-	m_targetControlMutex.unlock();
 	return reason;
+}
+
+
+DebugStopReason DebuggerController::LaunchAndWait(std::chrono::milliseconds timeout)
+{
+	return SubmitAndWait([this] { return LaunchAndWaitOnWorker(); }, timeout);
 }
 
 
@@ -354,8 +412,7 @@ bool DebuggerController::Attach()
 	if (!CanStartDebgging())
 		return false;
 
-	DbgRef<DebuggerController> self = this;
-	std::thread([self]() { self->AttachAndWait(); }).detach();
+	Submit([this] { AttachAndWaitOnWorker(); });
 	return true;
 }
 
@@ -363,7 +420,6 @@ bool DebuggerController::Attach()
 DebugStopReason DebuggerController::AttachAndWaitInternal()
 {
 	m_firstAttach = false;
-	m_userRequestedBreak = false;
 
 	DebuggerEvent event;
 	event.type = LaunchEventType;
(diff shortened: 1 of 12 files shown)


core/debuggercontroller.cpp
```diff
@@ -2692,54 +2942,33 @@ DebugStopReason DebuggerController::StopReason() const
...
 {
-	// Due to the nature of the wait, this mutex should NOT be allowed to be locked recursively.
-	// If this is a pause operation, do not try to lock the mutex -- it is mostly likely held by another thread
-	if ((operation != DebugAdapterPause) && (operation != DebugAdapterQuit) && (operation != DebugAdapterDetach))
+	// Invariant: ExecuteAdapterAndWait only ever runs on m_workerThread. The worker queue
+	// serializes all adapter operations, so the previous m_adapterMutex / m_adapterMutex2
+	// pair (which guarded against concurrent adapter access from multiple spawned threads)
+	// is no longer needed. The new Pause path bypasses this method entirely and calls
+	// m_adapter->BreakInto() out-of-band; everything else funnels through here on the worker.
+	assert(std::this_thread::get_id() == m_workerThreadId);
```
bdash: I'd recommend using BN_ASSERT or BN_RELEASE_ASSERT, depending on whether we want to validate this only in debug builds or also in our release builds. I'd lean towards the latter given that this check is cheap.

core/debuggercontroller.cpp
```diff
@@ -36,14 +37,26 @@ DebuggerController::DebuggerController(BinaryViewRef data): BinaryDataNotificati
...
+
+	m_workerShouldExit = false;
+	m_workerThread = std::thread([this] { WorkerThreadMain(); });
 }
 
 
 DebuggerController::~DebuggerController()
 {
+	{
+		std::lock_guard<std::mutex> lock(m_workQueueMutex);
```
bdash: I think there's an issue with the way mutexes and condition variables are used here that can lead to missed wakeups. Specifically, `DebuggerController::WaitForAdapterStop` uses `m_adapterStopMutex` with its condition variable, while this store occurs with `m_workQueueMutex` held. This means there's a race that can happen where `m_workerShouldExit` is set here but `DebuggerController::WaitForAdapterStop` does not see it.

This is related to the note at https://en.cppreference.com/cpp/thread/condition_variable that says:
> Even if the shared variable is atomic, it must be modified while owning the mutex to [correctly](https://stackoverflow.com/questions/38147825/) publish the modification to the waiting thread.

core/debuggercontroller.cpp
```diff
@@ -2043,6 +2268,92 @@ void DebuggerController::PostDebuggerEvent(const DebuggerEvent& event)
...
+
+void DebuggerController::RequestInterrupt()
+{
+	// Set the flag first so any in-flight op's silent-resume check observes the break
+	// request before the engine reports the resulting stop. Note: this path is purely
+	// out-of-band -- it does NOT call NotifyStopped. The in-flight worker op
+	// (GoAndWaitOnWorker etc.) is the sole notifier; it calls NotifyStopped on every
+	// genuine stop regardless of m_userRequestedBreak. m_userRequestedBreak only
+	// suppresses conditional-breakpoint auto-resume in ShouldSilentResumeAfterStop.
+	m_userRequestedBreak = true;
```
bdash: Is it safe to modify `m_userRequestedBreak` without taking any locks? It appears to potentially be read / written from different threads, so either needs to always be accessed with a lock held or to be `std::atomic<bool>`.

core/debuggercontroller.h
```diff
@@ -76,6 +77,17 @@ namespace BinaryNinjaDebugger {
 		};
 
 	private:
+		// m_adapter is the active debug adapter for this controller. Written exclusively
+		// from the worker thread (in CreateDebugAdapter); read both from the worker
```
bdash, line 90: Since this field is accessed by multiple threads, don't accesses need to be guarded by a mutex to avoid being data races?

core/debuggercontroller.cpp
```diff
@@ -60,6 +73,27 @@ DebuggerController::~DebuggerController()
 }
 
 
+void DebuggerController::WorkerThreadMain()
+{
+	m_workerThreadId = std::this_thread::get_id();
```
bdash: This is set on the worker thread but read from other threads without any synchronization, which is a data race. This needs some form of synchronization.

Another option would be to skip storing a thread ID, and instead use a thread-local variable that points to the `DebuggerController` instance when on the worker thread and is nullptr otherwise. To check whether code is running on the worker thread you'd then instead test `t_controllerOnWorker == this`. Since it is a thread-local variable there is no need for synchronization.

core/debuggercontroller.cpp
```diff
@@ -1859,8 +2027,24 @@ void DebuggerController::Destroy()
...
+// gotten around to running EventHandler. Now the controller's state is updated
+// happen-before the broadcast, and the dispatcher only fans out to external
+// (UI, plugin, scripting) consumers.
+//
+// Thread-safety: m_state's connection/execution status fields are simple atomic-ish
+// assignments. The other touched fields (m_lastIP, m_currentIP, m_exitCode, the
+// caches updated via UpdateCaches) are mutated under the worker's serialization
+// for ops the worker generates (TargetStopped, LaunchFailure, etc.) and under
+// the adapter event thread for spontaneous events (TargetExited, RegisterChanged,
+// ActiveThreadChanged). Concurrent mutation is unlikely in practice because the
```
bdash: I'm concerned by this reasoning. It seems to say that this code contains data races, but it's fine because they probably don't happen very often.

A data race is when two or more threads access the same memory location, with at least one of those accesses being a write, and there is no synchronization between the accesses (synchronization here is defined in terms of the [_happens-before_](https://en.cppreference.com/cpp/atomic/memory_order#Happens-before) relationship).

In C++, [if a data race occurs](https://en.cppreference.com/cpp/language/multithread#Data_races), the _behavior of the program is undefined_. The compiler is free to assume that data races do not happen and optimize the code on that basis. This makes it impossible to reason about what the code is actually doing.

core/debuggercontroller.h
```diff
@@ -201,6 +248,67 @@ namespace BinaryNinjaDebugger {
...
+			m_workQueueCv.notify_one();
+			return future;
+		}
+
+		// Submit a worker task and block the caller until the task completes (or the timeout
+		// elapses, in which case the engine is signaled to break and we still wait for the
+		// in-flight op to settle before returning). A timeout of milliseconds::max() means
+		// "wait forever" and bypasses wait_for entirely (avoids overflow inside the stdlib).
+		template<typename F>
+		auto SubmitAndWait(F&& f, std::chrono::milliseconds timeout)
```
bdash, line 330: What behavior does this have if run on the worker thread? `Submit` has an explicit case for it. Does this need one? If it's never expected to be called from the worker thread, an assert would serve to document that and catch any potential future misuse.

core/debuggercontroller.h
```diff
@@ -168,6 +194,27 @@ namespace BinaryNinjaDebugger {
...
+		DebugStopReason StepOverAndWaitOnWorker(BNFunctionGraphType il);
+		DebugStopReason StepOverReverseAndWaitOnWorker(BNFunctionGraphType il);
+		DebugStopReason StepReturnAndWaitOnWorker();
+		DebugStopReason StepReturnReverseAndWaitOnWorker();
+		DebugStopReason RunToAndWaitOnWorker(const std::vector<uint64_t>& remoteAddresses);
+		DebugStopReason RunToReverseAndWaitOnWorker(const std::vector<uint64_t>& remoteAddresses);
+		DebugStopReason RestartAndWaitOnWorker();
+		void DetachAndWaitOnWorker();
+		void QuitAndWaitOnWorker();
+		DebugStopReason PauseAndWaitOnWorker();
```
bdash: `PauseAndWaitOnWorker` appears to not be referenced anywhere.

xusheng6:
Hi @bdash thanks for the review! I forgot to mention that this PR mainly means to be clean up the lifetime + control-flow issues, so that I can deal with the data races in a follow-up PR. I did a bad job in communicating this, and I guess you could not have known that because the last commit I added did appear to trying to handle the data races. I know that one is not sufficient yet -- I was just testing the general visibility of the idea, and data races will be addressed in a future PR

I will go ahead and address the new issues you mention that are introduced by the new code, and have you look at it again

core/debuggercontroller.cpp
xusheng6: this boolean is a leftover from the old code. It is no longer needed in the new design, I will delete it

core/debuggercontroller.h
xusheng6: After looking at the code closer, I have decided to make both of them NOT re-entrant. There is also no place where they are called in a reentrant way.

Just for the record -- Submit can be made re-entrant safe, and SubmitAndWait, I believe, cannot

core/debuggercontroller.h
xusheng6: fixed

core/debuggercontroller.cpp
xusheng6: fixed!

core/debuggercontroller.cpp
xusheng6: I changed it to 

```
	{
		std::scoped_lock shutdownLock(m_workQueueMutex, m_adapterStopMutex);
		m_workerShouldExit = true;
	}
```

which I believe should fix the issue

core/debuggercontroller.cpp
xusheng6: I adopted the suggested method

core/debuggercontroller.h
xusheng6: yeah data races are not the focus of this PR, they will need a separate one to address

core/debuggercontroller.cpp
```diff
@@ -2036,14 +2033,16 @@ void DebuggerController::Destroy()
...
-// caches updated via UpdateCaches) are mutated under the worker's serialization
-// for ops the worker generates (TargetStopped, LaunchFailure, etc.) and under
-// the adapter event thread for spontaneous events (TargetExited, RegisterChanged,
-// ActiveThreadChanged). Concurrent mutation is unlikely in practice because the
-// engine produces one event at a time, but if it becomes a problem the mutex is
-// trivial to add.
+// Thread-safety: this body touches m_state's connection/execution status, plus
+// m_lastIP / m_currentIP / m_exitCode on the controller -- all plain (non-atomic,
+// unlocked) fields read from other threads (UI render layer, file accessor) without
+// synchronization. Those are pre-existing data races reported in #1091; this PR
```
bdash, line 2000: The remainder of this comment is in terms of "this PR" and "this commit", which won't make much sense once the change is merged.

I'd probably just end this with "Fixing these data races is tracked in #1091".

bdash (approved):

PR #1095, opened 2026-05-29
Fix some leaks

bdash:
I found these while working on https://github.com/Vector35/binaryninja/pull/1503, but then forgot I had them sitting in the debugger submodule 🙃


xusheng6:
Thanks @bdash

PR #1106, opened 2026-06-28
Resolve strdup memory leak with GetPathBaseName

SmoothHacker:
Whenever `DebugModule::GetPathBaseName` and `ModuleNameAndOffset::GetPathBaseName` is called a memory leak the size of the c string version of the path parameter is created because of `strdup()`. The fix involves using std::filesystem. We create a std::filesystem::path object and then return the steam of that path.

### DIFF
--- a/core/debugadapter.cpp
+++ b/core/debugadapter.cpp
@@ -16,12 +16,10 @@ limitations under the License.
 
 #include <binaryninjacore.h>
 #include <binaryninjaapi.h>
+#include <filesystem>
 #include <lowlevelilinstruction.h>
 #include <mediumlevelilinstruction.h>
 #include <highlevelilinstruction.h>
-#ifndef WIN32
-	#include "libgen.h"
-#endif
 #include "debugadapter.h"
 #include "debuggercontroller.h"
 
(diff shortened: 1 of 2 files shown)


xusheng6:
Thanks for fixing the `strdup` leak — that's a real leak worth removing. 👍

I pushed a one-line follow-up to your branch (627f0fa) changing `.stem()` to `.filename().string()`. Wanted to explain the reasoning:

**`stem()` and `basename()` aren't equivalent.** `std::filesystem::path::stem()` returns the filename with the *final extension stripped*, whereas the POSIX `basename()` you're replacing — and the Windows `_splitpath_s` branch, which returns `baseName + ext` — both keep the full final path component *including* the extension. So `stem()` silently changed the non-Windows behavior:

| input | old `basename()` / Windows branch | `stem()` | `filename()` |
|---|---|---|---|
| `/usr/lib/libc.so.6` | `libc.so.6` | `libc.so` ❌ | `libc.so.6` ✅ |
| `/usr/lib/libssl.so.1.1` | `libssl.so.1.1` | `libssl.so.1` ❌ | `libssl.so.1.1` ✅ |
| `/lib/foo.dll` | `foo.dll` | `foo` ❌ | `foo.dll` ✅ |

`GetPathBaseName` is used only for module-name matching (`IsSameBaseModule`, `StringsEqual`, and setting `m_short_name` on Windows), so stripping the extension would have (a) made Linux/macOS inconsistent with the Windows branch, and (b) risked false matches like `libfoo.so` vs `libfoo.dylib` both collapsing to `libfoo`. `filename()` returns the full final component and matches the original semantics exactly.

I also added the explicit `.string()` since `filename()` returns a `std::filesystem::path`; the implicit conversion to `std::string` works on POSIX but being explicit is clearer.

The extra includes and the leak fix itself look good. One optional suggestion: a small unit test for `GetPathBaseName` (a versioned `.so`, a `.dll`, and an extensionless name) would guard this behavior going forward, especially now that the repo builds test binaries.

xusheng6:
Thanks! @SmoothHacker

Issue #96, opened 2020-02-20
Read memory map from the backend


CouleeApps:
Currently the only information we have on segments is their binary path and start address. Because of this there is no way to determine their lengths and as a result, we don't know a lot of metadata about addresses (eg rwx). Also will fix Vector35/debugger#27 if we can create segments for only the mapped memory

xusheng6:
We not only need to read information about the mapped modules, we also need stacks, heaps, etc. 

Right now, since we are reporting the entire memory space as readable, the string search will try to read all of the bytes of the x86/x64 memory space, which is 0xffffffff or 0xffffffffffffffff bytes, which never finishes

xusheng6:
## Design Notes (from code review of x64dbg + internal discussion)

### How x64dbg handles memory maps
- **Full refresh, not incremental**: Every `MemUpdateMap()` clears the entire map and re-queries via `VirtualQueryEx()` loop. Typically 100-500 regions, takes ~1-10ms.
- **Background polling**: A dedicated thread refreshes the memory map every 2 seconds while debugging.
- **Smart lazy validation on stop**: When the debugger pauses, it checks if the current IP is already in the cached map. If yes → async refresh. If no → forced sync refresh.
- **Event-driven**: DLL load/unload triggers an async memory map refresh.

### Proposed design for our debugger

**Adapter capability**: Add `DebugAdapterSupportMemoryMap` to the `DebugAdapterCapacity` enum. Adapters that support it return `true` from `SupportFeature()`. Adapters that don't support it fall back to the current blanket memory region.

**Adapters that can support it**:
- **LLDB**: `SBProcess::GetMemoryRegions()` returns `SBMemoryRegionInfoList` with base, end, rwx permissions, name, mapped status.
- **GDB (RSP)**: `qMemoryRegionInfo:<hex-addr>` packet returns start, size, permissions. Walk the address space by jumping to `start+size`.
- **Windows native / DbgEng**: `VirtualQueryEx(hProcess, ...)` — both already have the process handle.

**One accessor, multiple named regions**: Instead of the current blanket `AddRemoteMemoryRegion("debugger", 0, accessor)` covering the full 64-bit space, add individual regions per memory mapping. The same shared `DebuggerFileAccessor` is reused — it already forwards reads to the backend regardless of address. BinaryNinja then knows which address ranges are valid and their permissions.

**Fallback**: If the adapter doesn't support memory map queries, keep the current blanket region behavior. No regression.

[...]

Issue #27, opened 2021-11-26
Test whether break in works


xusheng6:
Need to test when the target is running, whether break in actually works -- and the various components are properly updated

xusheng6:
It does work, though we have to skip the waiting queue in the QueuedAdapter. Same with the Python implementation. The problem will be resolved once we switch to fully event-driven

PR #1113, opened 2026-07-06
Add memory map to the debugger backend (#96)

xusheng6:
## Summary

First step toward #96 (and #27): expose the target's **memory map** — every mapped region of the virtual address space with its permissions — through the debugger, mirroring the existing *modules* feature end to end.

A memory region is deliberately a **separate abstraction** from a module, not an extension of it:

- a module (one entry per loaded binary) typically spans several regions with **different permissions** (`.text` r-x, `.data` rw-, …);
- many regions — **stack, heap, anonymous `mmap`** — belong to no module at all;
- on macOS the relationship is many-to-one *both ways* (the dyld shared cache packs many modules' `__TEXT` into one region and shares a single `__LINKEDIT`), so neither can be cleanly derived from the other.

So this adds a new `DebugMemoryRegion` type and a new `GetMemoryMap()` adapter API rather than overloading `DebugModule`.

## What's included

| Layer | Change |
|---|---|
| **core** | `DebugMemoryRegion` struct; `DebugAdapter::GetMemoryMap()` (default-empty so adapters opt in incrementally); `DebuggerMemoryMap` state cache (marked dirty on every stop, fetched lazily); `DebuggerController::GetMemoryMap()` |
| **adapters** | **LLDB** via `SBProcess::GetMemoryRegions()`; **GDB** via `/proc/[pid]/maps`, parsed per region (with rwx/shared) instead of aggregated by path the way `GetModuleList` does |
| **FFI** | `BNDebugMemoryRegion` + `BNDebuggerGetMemoryMap` / `BNDebuggerFreeMemoryRegions` |
| **C++ API** | `DebuggerController::GetMemoryMap()` |
| **Python** | `DebugMemoryRegion` + `controller.memory_map` |

`DebugMemoryRegion` carries `{start, size, name, read, write, execute, shared}`, where `name` is a backing file path, a pseudo-name like `[stack]`/`[heap]`, or empty for anonymous mappings.

## Scope / deliberately out of scope

This is the **data foundation only**. It does **not** change the debugger's live `BinaryView`: the flat remote region added in `CreateDebuggerBinaryView` still backs reads, so readability is unaffected on every platform.

The follow-up — replacing that flat region with **per-region segments** (real lengths + rwx permissions), which is the behavioral goal of #96/#27 — is intentionally a separate PR. Reads today go straight to the adapter (`DebuggerFileAccessor::Read` → `ReadMemory`) and are **not** gated by segments; the flip to per-region backing should land only after the map is validated as complete on each platform, so that no byte that is actually readable becomes unreadable. Landing the map first makes that validation possible (open the panel / call `controller.memory_map` and compare against real readability).

Adapters other than LLDB/GDB (DbgEng, etc.) inherit the default-empty `GetMemoryMap()` and return an empty map until implemented — no behavior change for them.

## Testing

- Builds and links cleanly (`ninja debuggercore debuggerapi`) on macOS.
- Python modules byte-compile.
- Manual verification of the region data on live targets (LLDB/GDB) is the main thing reviewers should exercise; that's also the validation gate before the per-region-segment follow-up.

### DIFF
--- a/api/debuggerapi.h
+++ b/api/debuggerapi.h
@@ -349,6 +349,17 @@ namespace BinaryNinjaDebuggerAPI {
 		static std::string GetPathBaseName(const std::string& path);
 	};
 
+	struct DebugMemoryRegion
+	{
+		std::uintptr_t m_start {};
+		std::size_t m_size {};
+		std::string m_name {};
+		bool m_read {};
+		bool m_write {};
+		bool m_execute {};
+		bool m_shared {};
+	};
+
 
 	struct DebugRegister
 	{
@@ -686,6 +697,7 @@ namespace BinaryNinjaDebuggerAPI {
 		bool ResumeThread(std::uint32_t tid);
 
 		std::vector<DebugModule> GetModules();
+		std::vector<DebugMemoryRegion> GetMemoryMap();
 		std::vector<DebugRegister> GetRegisters();
 		intx::uint512 GetRegisterValue(const std::string& name);
 		bool SetRegisterValue(const std::string& name, const intx::uint512& value);
--- a/api/debuggercontroller.cpp
+++ b/api/debuggercontroller.cpp
@@ -275,6 +275,31 @@ std::vector<DebugModule> DebuggerController::GetModules()
 }
 
 
+std::vector<DebugMemoryRegion> DebuggerController::GetMemoryMap()
+{
+	size_t count;
+	BNDebugMemoryRegion* regions = BNDebuggerGetMemoryMap(m_object, &count);
+
+	vector<DebugMemoryRegion> result;
+	result.reserve(count);
+	for (size_t i = 0; i < count; i++)
+	{
+		DebugMemoryRegion region;
+		region.m_start = regions[i].m_start;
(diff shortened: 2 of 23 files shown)


xusheng6:
Added a **Debugger Memory Map** sidebar widget (commit c9416c0) so the map can be exercised in the UI, not just via `controller.memory_map`.

- Mirrors the Debugger Modules panel: columns **Start / End / Size / Permissions / Name** (permissions shown as e.g. `r-xp` / `rw-s`).
- Refreshes on the target-stopped event; supports filtering, Copy / Copy All, and double-click / context-menu **Jump To Start / End**.
- Appears in the right sidebar (`RightBottom`, hidden when there's no active debug session), same as the Modules panel.

**How to test:** launch/attach a target (LLDB or GDB), pause it, open the *Debugger Memory Map* sidebar icon. Compare against `/proc/<pid>/maps` (Linux) or `vmmap <pid>` (macOS) to sanity-check completeness — that comparison is also the validation gate for the per-region-segment follow-up. Other adapters (DbgEng, etc.) show an empty panel until `GetMemoryMap()` is implemented for them.

PR #1122, opened 2026-07-20
Fix memory reported unreadable during TTD when readable region is smaller than a cache block

### LINKED ISSUE #725: Memory bytes reported as unreadable though it is indeed readable during TTD
While using dbgeng ttd adapter to debug a recorded trace, I observe that sometimes the memory bytes can appear unreadable in the linear view, but it is indeed readable if I directly run a windbg command:

Image

xusheng6:
Fixes #725.

During TTD, memory readability is tracked at the granularity of what the recorded execution actually touched, so a readable region can be **smaller than the `0x100` chunk** the debugger reads and can **start at a non-`0x100`-aligned address**. Two bugs caused such memory to be reported as entirely unreadable in the linear view (even though `dx`/windbg could read it):

### 1. Adapter discarded partial reads (`core/adapters/dbgengadapter.cpp`)
`DbgEngAdapter::ReadMemory` required `bytesRead == size`, so a valid partial read (e.g. `0x20` of a requested `0x100` bytes) was thrown away as a total failure. This went unnoticed in live debugging because the OS allocator hands out page-sized regions. It now returns the bytes actually read whenever `ReadVirtual` succeeds and reads ≥ 1 byte — the contract other adapters (e.g. GDB) already honor.

### 2. Cache poisoned by holes before the requested address (`core/debuggerstate.{h,cpp}`)
`DebuggerMemory::ReadMemory` rounded every read **down to a `0x100` block boundary** and read the whole block from its base. When the base fell in an unreadable hole, the entire block was marked `FailedToRead`, shadowing the readable bytes that follow the hole.

The cache now:
- anchors each backend read at the **exact requested address** (no round-down), so an unreadable region *before* the requested address is never touched; and
- records read failures at **single-byte granularity**, so a hole never shadows a readable region immediately after it.

`ReadBlock` is replaced by `ReadRun`; the cache stores variable-length runs (readable bytes or one-byte hole markers) keyed by start address. This fix lives in the **generic cache layer**, so it applies to every adapter, not just DbgEng.

### Reproduction
Reproduced with a TTD trace where a separate process `VirtualAllocEx` + `WriteProcessMemory`s a small, misaligned `0x20`-byte buffer into the traced target (so the target executes no write instruction on that page and TTD never page-snapshots it); the target's byte-granular reads then leave a sub-`0x100`, non-aligned readable window. Confirmed via dbgeng that the window is `0x40` bytes at a non-`0x100`-aligned address, and that `read_memory` returns `b''` on a pre-fix build vs the bytes on the fixed build. (Test source/binaries not included in this PR.)

### DIFF
--- a/core/debuggerstate.cpp
+++ b/core/debuggerstate.cpp
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+#include <algorithm>
 #include <chrono>
 #include <thread>
 #include <utility>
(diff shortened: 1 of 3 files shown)


core/debuggerstate.cpp
```diff
@@ -1396,32 +1413,30 @@ DataBuffer DebuggerMemory::ReadMemory(uint64_t offset, size_t len)
...
+		pos += take;
+
+		if (take < run.GetLength())
+			// The run more than covered the rest of the request; we are done
+			break;
+
+		// Otherwise the run may have been cut short by a hole at `pos`; the next ReadRun(pos) will report it and
+		// terminate the loop (also caching the hole marker for that byte).
 	}
 	BN_RELEASE_ASSERT(result.GetLength() <= len);
```
plafosse, line 1441: lets remove this assert or alt least make it a BN_ASSERT

core/debuggerstate.cpp
```diff
@@ -1396,32 +1413,30 @@ DataBuffer DebuggerMemory::ReadMemory(uint64_t offset, size_t len)
 
 	DataBuffer result;
```
plafosse, line 1414: should probably reserve `len`

core/debuggerstate.h
```diff
@@ -227,7 +231,9 @@ namespace BinaryNinjaDebugger {
 		DebuggerMemory(DebuggerState* state);
 
 		void MarkDirty();
-		DataBuffer ReadBlock(uint64_t block);
+		// Reads the readable run starting exactly at `address` (up to 0x100 bytes), or an empty buffer if
+		// `address` itself is unreadable. The address is NOT rounded down to a cache-block boundary.
+		DataBuffer ReadRun(uint64_t address);
```
plafosse: Change to ReadCachedBlock or something more descriptive. Also make the 0x100 bytes into a constexpr

plafosse (approved):
minor nit about function name and ensure the asserts are removed.

core/debuggerstate.cpp
xusheng6: This assert just asserts something that should always be true -- it does not assert on `result.GetLength() == len` xD

core/debuggerstate.cpp
xusheng6: See https://github.com/Vector35/debugger/pull/1122#discussion_r3624809653, so I am not gonna do it here

PR #1125, opened 2026-07-20
Add TTD go to previous/next register write (#1123)

### LINKED ISSUE #1123: Add suppport for "go to previous/next register write" in Windbg TTD
### Summary

During TTD analysis, one of the most common questions is "where was this
register last written?" WinDbg exposes this directly:

- `!tt br <reg>` — travel to the previous/next position where a register changes value
- `dx @$curthread.TTD.PrevRegisterWrite("<reg>" [, <value>])` and the matching
  `NextRegisterWrite(...)` — return the position/value and can be seeked to

Today this requires dropping into the debugger console and typing the DDM
expression by hand. It'd be a nice workflow win to expose it as a first-class
action in the debugger UI.

### Proposed UX

- Right-click a register in the Registers view → **Go to previous write** /
  **Go to next write**, which seeks the TTD position.
- Expose an equivalent in the debugger Python API (a helper wrapping
  `PrevRegisterWrite`/`NextRegisterWrite` + seek) so it's scriptable from BN.

### Backend note

The primitive already exists in dbgeng — register-change navigation landed in
TTD ~1.11.x (April 2025 WinDbg release), via both `!tt br` and the
`Prev/NextRegisterWrite` DDM methods, hanging off `@$curthread` (per-thread).
So this is primarily plumbing a UI/API action to an existing capability.

### Caveat worth documenting

These key on the register *changing value*, not on every architectural write —
a write of the same value the register already held won't be found. Worth
noting in the tooltip/docstring so it doesn't surprise anyone doing data-flow
work.

xusheng6:
Fixes #1123.

Adds "go to previous/next register write" for TTD — locate and time-travel to where a register's value last changed, via the dbgeng `PrevRegisterWrite`/`NextRegisterWrite` data-model methods.

- Right-click a register in the **Registers** widget → *Go to Previous/Next Write*
- Right-click a register token in the **graph/linear/hex** view → *Debugger → Go to Previous/Next Register Write*
- C++/Python APIs: `get_ttd_prev/next_register_write()`
- Docs updated (`dbgeng-ttd.md`, `ttd-python-api.md`)

Note: TTD reports register value *changes*, not every architectural write.

### DIFF
--- a/api/debuggerapi.h
+++ b/api/debuggerapi.h
@@ -902,8 +902,8 @@ namespace BinaryNinjaDebuggerAPI {
 		bool SetTTDPosition(const TTDPosition& position);
 		std::pair<bool, TTDMemoryEvent> GetTTDNextMemoryAccess(uint64_t address, uint64_t size, TTDMemoryAccessType accessType);
 		std::pair<bool, TTDMemoryEvent> GetTTDPrevMemoryAccess(uint64_t address, uint64_t size, TTDMemoryAccessType accessType);
-		std::pair<bool, TTDRegisterWriteEvent> GetTTDNextRegisterWrite(const std::string& reg);
-		std::pair<bool, TTDRegisterWriteEvent> GetTTDPrevRegisterWrite(const std::string& reg);
+		std::optional<TTDRegisterWriteEvent> GetTTDNextRegisterWrite(const std::string& reg);
+		std::optional<TTDRegisterWriteEvent> GetTTDPrevRegisterWrite(const std::string& reg);
 
 		// TTD Position History Navigation
 		bool TTDNavigateBack();
--- a/api/debuggerapi.h
+++ b/api/debuggerapi.h
@@ -531,6 +531,18 @@ namespace BinaryNinjaDebuggerAPI {
 		TTDMemoryEvent() : threadId(0), uniqueThreadId(0), accessType(TTDMemoryRead), address(0), size(0), memoryAddress(0), instructionAddress(0), value(0) {}
 	};
 
+	struct TTDRegisterWriteEvent
+	{
+		std::string reg;               // Register name that was queried (e.g. "rax")
+		TTDPosition position;          // Position at which the register value changed
+		TTDPosition originalPosition;  // Position from which the query was issued
+		uint64_t value;                // Value the register was changed to
+		uint64_t originalValue;        // Value the register held before the change
+		uint32_t uniqueThreadId;       // Unique thread identifier that performed the write
+
+		TTDRegisterWriteEvent() : value(0), originalValue(0), uniqueThreadId(0) {}
+	};
+
 	struct TTDPositionRangeIndexedMemoryEvent{
 		TTDPosition position;				// Position of the memory event
 		uint32_t threadId;					// Thread ID that performed the access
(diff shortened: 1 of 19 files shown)


api/python/debuggercontroller.py
```diff
@@ -892,6 +892,60 @@ def __repr__(self):
         return f"<TTDMemoryEvent: {self.event_type} @ {self.address:#x}, thread {self.thread_id}>"
 
 
+class TTDRegisterWriteEvent:
```
plafosse, line 895: Looks like this should be implemented as a frozen dataclass

api/python/debuggercontroller.py
```diff
@@ -3030,6 +3084,66 @@ def get_ttd_prev_memory_access(self, address: int, size: int, access_type = Debu
             access_type=result.accessType
         )
 
+    def get_ttd_next_register_write(self, reg: str) -> Optional[TTDRegisterWriteEvent]:
```
plafosse, line 3087: just a comment but generally we should avoid using a native `str` for what should a typedef'd  type in architecture.py we use

```
RegisterName = NewType('RegisterName', str)
```

api/debuggerapi.h
```diff
@@ -890,6 +902,8 @@ namespace BinaryNinjaDebuggerAPI {
 		bool SetTTDPosition(const TTDPosition& position);
 		std::pair<bool, TTDMemoryEvent> GetTTDNextMemoryAccess(uint64_t address, uint64_t size, TTDMemoryAccessType accessType);
 		std::pair<bool, TTDMemoryEvent> GetTTDPrevMemoryAccess(uint64_t address, uint64_t size, TTDMemoryAccessType accessType);
+		std::pair<bool, TTDRegisterWriteEvent> GetTTDNextRegisterWrite(const std::string& reg);
```
plafosse: This shoudl probably be a string_view instead

core/adapters/dbgengttdadapter.cpp
```diff
@@ -1481,6 +1539,131 @@ std::pair<bool, TTDMemoryEvent> DbgEngTTDAdapter::ParseSingleTTDMemoryObject(con
 }
 
 
+std::pair<bool, TTDRegisterWriteEvent> DbgEngTTDAdapter::ParseSingleTTDRegisterWriteObject(const std::string& expression, const std::string& reg)
```
plafosse: I believe this would be better as a std::string (non-const-ref) and do a std::move into wExpression

core/adapters/dbgengttdadapter.cpp
```diff
@@ -1481,6 +1539,131 @@ std::pair<bool, TTDMemoryEvent> DbgEngTTDAdapter::ParseSingleTTDMemoryObject(con
...
+			nullptr,
+			result.GetAddressOf(),
+			metadata.GetAddressOf()
+		);
+
+		if (FAILED(hr))
+		{
+			// A failed evaluation typically means there is no previous/next write of this
+			// register within the trace, which is a normal "not found" outcome.
+			LogInfo("No register write found for expression '%s' (0x%08x)", expression.c_str(), hr);
```
plafosse, line 1578: Avoid using Log* without a logger. And prefer Log*F variants, as their faster and easier to read and more featurefull.

api/debuggerapi.h
plafosse: these would be better as std::optional rather than pair<bool, ...>

plafosse (approved):
I'm approving this contingent on fixing the std::pair<bool, ...> thing. The other comments are really more things that are more wholistic and we need some light mechanical refactors on.

xusheng6:
Created https://github.com/Vector35/debugger/issues/1131 to track the chores

PR #1128, opened 2026-07-20
Switch WinDbg installer to miniz and extract inner MSIX from memory (#1127)

### LINKED ISSUE #1127: WinDbg installer: redundant intermediate disk extraction and no feedback during temp cleanup
## Summary

The WinDbg/TTD installer used a hand-rolled ZIP/inflate parser and materialized a large intermediate file to disk, so the install appeared to hang on the "Verifying installation..." step.

## Background

For a current WinDbg package the installer touched roughly **1.9 GB** of temporary disk I/O. The inner `windbg_win-x64.msix` (381 MB) was written to a temp file only to be reopened and read straight back, then deleted.

## Symptom: "Verifying installation..." appears stuck

`CheckInstallation()` itself is instant (six `fs::exists` checks). The perceived hang was the `CleanupTempFiles()` call that ran right after it. Measured: deleting the freshly-written 381 MB inner `.msix` took **~68 s** — not because the raw delete is slow (that's ~85 ms) but because **antivirus scans the just-written archive synchronously on delete**.

## Resolution (landed in #1128)

- **Replaced the homegrown ZIP/inflate parser** (`vendor/minizip-ng`, ~45 KB of bespoke code — despite the name, *not* upstream minizip-ng) with the vendored, well-audited **miniz** amalgamation. Also removes a bespoke parser from a supply-chain-sensitive path and fixes the old reader's 2 GB (`long`) offset limit.
- **Extract the inner package directly from memory** (`mz_zip_reader_init_mem`) into the install dir — no intermediate `.msix` file is written or later deleted.
- Gave the cleanup step its own progress message.

**Measured (real 1.13 GB package):** whole install ~87 s → ~17 s; cleanup ~68 s → 0 ms (nothing to delete). The remaining ~17 s is the actual writing of 1446 payload files.

### Follow-up (not in #1128)

In the real download path the 1.13 GB `.msixbundle` itself is still deleted at the end and could hit the same antivirus scan-on-delete cost. Worth measuring separately; the inner-package temp (the ~68 s case) is resolved.

xusheng6:
Fixes #1127.

## Summary

The installer used a hand-rolled ZIP/inflate parser (`vendor/minizip-ng`, ~45 KB of bespoke code — despite the folder name, it is **not** upstream minizip-ng) sitting directly in the download path, and it extracted the inner `windbg_win-x64.msix` to a **temporary file** before extracting its contents. That temp file was the real cause of the slow "cleanup": deleting the freshly-written 381 MB `.msix` took ~68 s because antivirus scans the archive synchronously on delete.

## Changes

- **Replace the homegrown parser with vendored miniz 3.0.2** (`vendor/miniz`, amalgamated `miniz.c`/`miniz.h` + `LICENSE`) and delete `vendor/minizip-ng` entirely. miniz is a widely-used, well-audited single-file library bundling both ZIP reading and inflate (no separate zlib backend needed). It also uses 64-bit file offsets, fixing the old reader's 2 GB (`long`) limit.
- **Extract the inner package from memory** (`mz_zip_reader_init_mem`) straight into the install dir — no intermediate `.msix` is written or later deleted.
- Give the cleanup step its own "Cleaning up temporary files..." progress message (it no longer looks stuck on "Verifying installation...").
- Add a `--bundle <file>` / `--verbose` option to the installer CLI to install from an already-downloaded bundle (kept, not deleted) with timing logs — used to measure the changes below.

## Measured (via `--bundle` on a real 1.13 GB package)

| | Before | After |
|---|---|---|
| Inner MSIX temp write | 381 MB to disk | none |
| Cleanup (delete) | ~68,000 ms | 0 ms |
| Whole install (no download) | ~87 s | ~17 s |

1446 payload files extracted; all required files (`dbgeng.dll`, `TTD.exe`, …) present. The remaining ~17 s is the actual writing of those files.

## Security note

This removes a bespoke ZIP/inflate parser from a supply-chain-sensitive download path in favor of a maintained, widely-reviewed library — motivation for the switch.


xusheng6:
We want to see if we can use binja's container file support to do the extraction

xusheng6:
> We want to see if we can use binja's container file support to do the extraction

@plafosse this cannot use the binja's zip support -- it is meant to be executed as a standalone executable, and it does not link with binja (or it would not work for the free version). As such, I think miniz is a reasonable choice

xusheng6:
Superseded by #1202, which uses the native Windows App Packaging API.

PR #1133, opened 2026-07-22
Skip the adapter settings dialog on subsequent debug sessions

xusheng6:
## Skip the adapter settings dialog on subsequent debug sessions

Adds a **"Use same settings next time"** checkbox to the Debug Adapter Settings dialog. When checked, the dialog is not shown again the next time you start the debugger with the same operation, so repeated launches/attaches don't re-prompt for settings.

Supersedes #1064 — built on top of @3rdit's original work (credited as co-author). Rebased onto latest `dev`.

### Behavior

The preference is a single flag combined with a record of the **last debug-start operation** (launch / attach / connect / connect-to-debug-server). The dialog is shown when:

- the operation differs from the last one you ran, **or**
- the "use same settings" preference asks for it (i.e. the box was left unchecked).

This avoids a footgun a naive single flag would have: choosing "use same settings" while **launching** must not silently suppress the dialog when you later switch to **attach** — an operation you never configured. Because switching operations always re-prompts, suppression is always explicit and per-kind, while the only inherited direction ("show again") is harmless.

The checkbox reflects and writes this one preference. It also appears in the generic **Debug Adapter Settings** dialog (opened from the menu), which is the entry point for re-enabling a dialog that has been suppressed for an operation.

The preference is in-memory (per debugger session), matching the existing `IsFirstLaunch`/`IsFirstAttach` flags.

### Changes

- **core**: replace the four per-operation "show next time" flags with a single `m_showAdapterSettingsNextTime` plus `m_lastDebugStartOperation`, recorded when each session starts. New `ShouldShowAdapterSettingsFor{Launch,Attach,Connect,ConnectToDebugServer}()` decision helpers and a single `ShowAdapterSettingsNextTime()` getter / `SetShowAdapterSettingsNextTime()` setter.
- **ffi / api**: collapse the eight exported functions to six (four per-op decisions + getter + setter).
- **ui**: dialog gates simplified to `ShouldShowAdapterSettingsFor<Op>()`; checkbox writes the single preference and now also shows in the generic settings dialog (applied live, since that dialog has no OK/Cancel).

### Testing

Manually verified: first launch prompts; a second launch with the box checked skips the dialog; switching to attach re-prompts; unchecking the box in the generic settings dialog re-enables prompting.

### DIFF
--- a/api/debuggerapi.h
+++ b/api/debuggerapi.h
@@ -890,6 +890,16 @@ namespace BinaryNinjaDebuggerAPI {
 		bool IsFirstConnectToDebugServer();
 		bool IsFirstAttach();
 
+		bool ShouldShowAdapterSettingsNextLaunch();
+		bool ShouldShowAdapterSettingsNextAttach();
+		bool ShouldShowAdapterSettingsNextConnect();
+		bool ShouldShowAdapterSettingsNextConnectToDebugServer();
+
+		void SetShowAdapterSettingsNextLaunch(bool value);
+		void SetShowAdapterSettingsNextAttach(bool value);
+		void SetShowAdapterSettingsNextConnect(bool value);
+		void SetShowAdapterSettingsNextConnectToDebugServer(bool value);
+
 		bool IsTTD();
 
 		// TTD Memory Analysis Methods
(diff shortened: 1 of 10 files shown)


xusheng6:
This is from a user PR (https://github.com/Vector35/debugger/pull/1064), I reviewed it, made some changes to it, and merged it

PR #1138, opened 2026-07-23
Navigate double-clicked sidebar addresses by content type

### LINKED ISSUE #1134: Double clicking a register value should also navigate to it in a new pane
Same as when we double click a register token in graph/linear view

xusheng6:
Double-clicking an address in a debugger sidebar (register value, breakpoint, module, memory region, stack frame, debugger-info entry, or a TTD address column) now chooses the pane by content type:

- **Same kind** as the current pane (both code, or both data) → navigate the current pane in place.
- **Different kind** (code ↔ data) → open in the other pane, so following a data pointer no longer hijacks the code view. Because data targets then match the (data) companion pane, repeated double-clicks stop drifting across panes.

A right-click **"Navigate in Current Pane"** action forces navigation into the focused pane (added to Stack Trace, Debugger Info, and the TTD Calls/Memory/Events widgets; the other widgets already had an equivalent same-pane action).

The graph/linear token double-click and the TTD time-travel columns are intentionally left unchanged.

Fixes https://github.com/Vector35/debugger/issues/1134

### DIFF
--- a/ui/debuggerinfowidget.cpp
+++ b/ui/debuggerinfowidget.cpp
@@ -17,10 +17,13 @@ limitations under the License.
 #include <QPainter>
 #include <QHeaderView>
 #include <QGuiApplication>
+#include <QMenu>
+#include <QContextMenuEvent>
 #include <QMimeData>
 #include <QClipboard>
 #include "ui.h"
 #include "debuggerinfowidget.h"
+#include "debuggeruicommon.h"
 #include "lowlevelilinstruction.h"
 #include "mediumlevelilinstruction.h"
 #include "highlevelilinstruction.h"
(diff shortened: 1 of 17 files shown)


ui/debuggeruicommon.cpp
```diff
@@ -69,3 +71,52 @@ bool ParseAddress(const QString& text, BinaryNinja::Ref<BinaryNinja::BinaryView>
...
+	if (!context)
+		return;
+
+	ViewFrame* frame = context->getCurrentViewFrame();
+	if (!frame)
+		return;
+
+	View* view = frame->getCurrentViewInterface();
+
+	// Decide same-pane vs other-pane by whether the target and the current view are the
```
fuzyll, line 94: This seems really fragile. Are we not able to know what the view is here? Also confused because some views show both code _and_ data. So, wouldn't this code potentially navigate the wrong pane in that situation?

For example, if you have Hex View open and Linear View open and you click an address inside of a function, Hex View can show "code", so I think this would navigate the Hex View and not the Linear View?

ui/ttdeventswidget.cpp
```diff
@@ -817,34 +826,55 @@ void TTDEventsQueryWidget::onCellDoubleClicked(int row, int column)
...
+			// kind of thing (code vs data) than the current pane shows (see NavigateToAddress,
+			// issue #1134). The Position (time-travel) column above keeps the current pane.
+			NavigateToAddress(this, m_data, address);
+			updateStatus(QString("Navigated to address %1").arg(cellText));
 		}
 	}
 }
 
+
+bool TTDEventsQueryWidget::addressForCell(int row, int column, uint64_t& addr)
```
fuzyll, line 844: My LLM says the following:

> This function does not recognize the specialized Exception Events Program Counter column. Its name does not contain the contiguous substring PC, and its physical index is not ExceptionPCColumn. Consequently, the new “Navigate in Current Pane” action is disabled for that address.

...and then followed up with:

> The same code treats column index ModuleAddressColumn as an address regardless of the specialized table schema. In Module Events, that physical index is Module Checksum, so showing that normally hidden column enables “Navigate in Current Pane” and interprets its checksum as an address. Column recognition needs to account for m_widgetType instead of reusing the AllEvents indices.

ui/debuggerinfowidget.cpp
```diff
@@ -935,25 +938,56 @@ void DebuggerInfoTable::updateFonts()
...
 
-	ViewFrame* frame = context->getCurrentViewFrame();
-	if (!frame)
+void DebuggerInfoTable::onDoubleClicked()
+{
+	uint64_t value = 0;
+	if (!selectedValue(value))
 		return;
 
+	// Navigate to the target, opening it in the other pane when it is a different kind
```
fuzyll, line 958: This comment is above every single call to `NavigateToAddress`. This is not a good idea. Not only do we already have this content within the function itself, but we also have it in the `.h` file. If the implementation ever changes, we have to update that information in two places _plus_ every single location that calls it. That's _way_ too much.

Just leave the header and function comments and remove all of the extra comments above every call to it.

ui/threadframes.cpp
```diff
@@ -849,30 +860,50 @@ void ThreadFramesWidget::onDoubleClicked()
...
+}
+
+
+bool ThreadFramesWidget::navigationAddressForSelection(uint64_t& addr)
+{
+	QModelIndexList sel = selectionModel()->selectedIndexes();
+	if (sel.empty())
+		return false;
+
+	const QModelIndex& index = sel[0];
```
fuzyll, line 879: You re-read the index here, but I thought `QTreeView::doubleClicked` would [already provide this](https://doc.qt.io/qt-6/qabstractitemview.html#doubleClicked)? Seems like we should be plumbing the `index` from that through to a function like this, rather than re-reading it? I believe this pattern is probably in other places, too, and I'm not sure if I'm just missing something or if this is doing work it really shouldn't be.

fuzyll (changes requested):
Some of this seems superfluous/confusing, and I am concerned about a possible usability issue with the pane-deciding code. The rest seems fairly straight-forward, though.

PR #1139, opened 2026-07-23
Refine double-click navigation while debugging

### LINKED ISSUE #1136: Double clicking a PossibleAddressToken always navigates to it in a new pane
But when the target is in the current function, the user probably just wants to navigate to it in the current pane

### LINKED ISSUE #1137: Debugger should also intercept the navagation of a data symbol token
Like when data_100003fa9 is double-clicked, it should navigate to it in the other pane

Image

xusheng6:
Two fixes to `OnTokenDoubleClicked`:

- **#1136**: Address literals resolving into the current function now navigate in place instead of always opening a second pane. Only out-of-function targets open in another pane.
- **#1137**: Data symbol (e.g. `data_100003fa9`) and inline string tokens now open in another pane while debugging, instead of falling through to default in-pane navigation.

Closes #1136
Closes #1137

### DIFF
--- a/ui/uinotification.cpp
+++ b/ui/uinotification.cpp
@@ -271,11 +271,18 @@ bool NotificationListener::OnTokenDoubleClicked(UIContext* context, ViewFrame* f
 	}
 	else if (token.type == DataSymbolToken || token.type == StringToken)
 	{
-		// A data symbol (e.g. data_100003fa9) or an inline string. The default behavior
+		// A data symbol (e.g. data_100003fa9) or a string reference. The default behavior
 		// navigates to it in the same view; while debugging we instead open it in another
-		// pane so the disassembly the user is looking at stays put. The referenced address is
-		// carried in the token's value (the address field is not populated for these tokens).
-		target = token.addrValid ? token.addr : token.token.value;
+		// pane so the disassembly the user is looking at stays put.
+
+		if ((token.type == StringToken) && (token.token.context != StringReferenceTokenContext)
+			&& (token.token.context != StringDataVariableTokenContext))
+			return false;
+
+		if (!token.addrValid)
+			return false;
+
+		target = token.addr;
 		haveTarget = true;
 	}
 	else if (token.type == RegisterToken)
@@ -341,6 +348,5 @@ bool NotificationListener::OnTokenDoubleClicked(UIContext* context, ViewFrame* f
 		return false;
 
 	// Open the target in another pane so the disassembly the user is looking at stays put.
-	view->navigateOnOtherPane(target);
-	return true;
+	return view->navigateOnOtherPane(target);
 }
--- a/ui/uinotification.cpp
+++ b/ui/uinotification.cpp
@@ -269,6 +269,15 @@ bool NotificationListener::OnTokenDoubleClicked(UIContext* context, ViewFrame* f
 		target = token.addr;
 		haveTarget = true;
 	}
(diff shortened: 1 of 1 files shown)


xusheng6:
Just wish to mention -- I somehow think that this should actually become the default behavior when the user is not debugging, or at least an option that can be enabled

ui/uinotification.cpp
```diff
@@ -246,9 +246,38 @@ bool NotificationListener::OnTokenDoubleClicked(UIContext* context, ViewFrame* f
...
+			{
+				if ((token.addr >= range.start) && (token.addr < range.end))
+					return false;
+			}
+		}
+
 		target = token.addr;
 		haveTarget = true;
 	}
+	else if (token.type == DataSymbolToken || token.type == StringToken)
```
fuzyll, line 272: I am reasonably certain this is too broad? In some other places where we deal with `StringToken`s, we also check the `token.context` to see what _kind_ of token it is. A `StringToken` that's `StringDisplayTokenContext` is, I believe, not really intended to be navigable in this way. For example, I think if you double-clicked on a calling convention string, you would wind up navigating to...I genuinely do not know where, with this code.

fuzyll (changes requested):
Only saw one thing I think we should change. I did not have a chance to test my theory about where this would show up because I did not build this myself, though, so I might be wrong.

ui/uinotification.cpp
xusheng6: Fixed in https://github.com/Vector35/debugger/pull/1139/commits/1038bd06613aaff919520240b4aa6e7680222018 by checking the type of the context

ui/uinotification.cpp
```diff
@@ -312,6 +348,5 @@ bool NotificationListener::OnTokenDoubleClicked(UIContext* context, ViewFrame* f
 		return false;
 
 	// Open the target in another pane so the disassembly the user is looking at stays put.
-	view->navigateOnOtherPane(target);
-	return true;
+	return view->navigateOnOtherPane(target);
```
fuzyll: Does this change the behavior in a way we didn't account for before? Before, we always returned `true`. Now, I'm guessing we'll return `false` if the navigation failed?

fuzyll (approved):
Approved, but we may want to double-check this one line to make sure it's the appropriate change. Probably is, just wasn't expecting the change, and didn't see it mentioned.

ui/uinotification.cpp
xusheng6: Yeah I think the debugger should not just eat the double click navigation when it cannot do it

xusheng6: Well I decide to test it out a bit more. It is also bad if the user double clicks something, expects a navigation, and the navigation does not happen, and something *else* happens. From this perspective, the debugger should just eat an attempted navigation even if it does not go anywhere

fuzyll: I'm not sure what the right answer is, and I'm fine with either answer, I just wanted to make sure this was an intentional change and not an LLM "improving" something without oversight.

xusheng6: yeah no worries, it is introduced by LLM, but I spotted it and think it is a good idea

xusheng6: I ended up deciding just let the debugger eat the event if it thinks the token should be navigable, but the navigation fails

xusheng6:
Thanks @fuzyll!

PR #1145, opened 2026-07-24
Reject a TTD launch with no trace specified (Fixes #910)

### LINKED ISSUE #910: prompt for trace when TTD backend selected
If the TTD backend is selected but a trace isn't specified, can we either prevent the settings dialog from being accepted, or pop up separately afterward? (suggestion from stream...)

xusheng6:
Fixes #910.

This PR fails the TTD launch attempt and gives a meaningful error message (rather than the cryptic OpenDumpFile failed with xxxxx)

The check are both added in the UI and the debug adapter

### DIFF
--- a/core/adapters/dbgengttdadapter.cpp
+++ b/core/adapters/dbgengttdadapter.cpp
@@ -31,7 +31,7 @@ bool DbgEngTTDAdapter::ExecuteWithArgsInternal(const std::string& path, const st
 	// HRESULT, so check it up front. The UI blocks this in the adapter settings dialog; this covers the
 	// headless and scripted paths, plus the launches where that dialog is skipped.
 	std::error_code ec;
-	if (tracePath.empty() || !std::filesystem::exists(tracePath, ec))
+	if (tracePath.empty() || !std::filesystem::is_regular_file(tracePath, ec))
 	{
 		DebuggerEvent event;
 		event.type = LaunchFailureEventType;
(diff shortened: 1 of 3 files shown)


ui/adaptersettings.cpp
```diff
@@ -158,8 +160,46 @@ QWidget* AdapterSettingsDialog::getWidgetForAdapter(const QString& adapter)
...
+	if (tracePath.empty())
+	{
+		QMessageBox::warning(this, "No Trace Specified",
+			"The TTD adapter replays a previously recorded trace, so a trace path is required.\n\n"
+			"Set \"Trace Path\" in the launch settings, or record a new trace first.");
+		return false;
+	}
+
+	std::error_code ec;
+	if (!std::filesystem::exists(tracePath, ec))
```
fuzyll: Need to check that this is a file (e.g. `std::filesystem::is_regular_file`) as well, I think?

core/adapters/dbgengttdadapter.cpp
```diff
@@ -27,6 +27,30 @@ bool DbgEngTTDAdapter::ExecuteWithArgsInternal(const std::string& path, const st
 	scope = SettingsResourceScope;
 	auto inputFile = adapterSettings->Get<std::string>("common.inputFile", data, &scope);
 
+	// A trace is not optional for replay, but a missing one only surfaces later as an opaque OpenDumpFile
+	// HRESULT, so check it up front. The UI blocks this in the adapter settings dialog; this covers the
+	// headless and scripted paths, plus the launches where that dialog is skipped.
+	std::error_code ec;
+	if (tracePath.empty() || !std::filesystem::exists(tracePath, ec))
```
fuzyll: Need to check that this is a file (e.g. `std::filesystem::is_regular_file`) as well, I think?

fuzyll (approved):
Looks good on static review, but unless something earlier rejects directories that I missed, I think just using `std::filesystem::exists` is not quite enough in two places.

ui/adaptersettings.cpp
xusheng6: Yeah we should check for regular files. I will fix it

xusheng6:
Merged with approval.

I noticed that we can make it one step further by checking the given trace file is not a valid trace file, e.g., by checking the magic header of the trace against the known good value, or at least reject it when a PE is selected. A common foot-gun is the user would incorrectly select the PE file instead of the trace file before launching. 

But we need to be sure to account for the possibility that MS would change the trace file magic without us knowing, so we would also need to have a setting to skip this check. So much said, we would better leave it as a future enhancement

PR #1147, opened 2026-07-24
Remember the last checked access types in the TTD next/prev memory access dialog (Fixes #1109)

### LINKED ISSUE #1109: TTD go to last/next memory access should save the last checked operation

xusheng6:
The dialog reset its Read/Write/Execute checkboxes on every open. Persist the selection with QSettings and restore it, falling back to the address-based guess only until the user picks something

### DIFF
--- a/ui/ttdmemorywidget.cpp
+++ b/ui/ttdmemorywidget.cpp
@@ -29,6 +29,7 @@ limitations under the License.
 #include <QToolButton>
 #include <QPropertyAnimation>
 #include <QFrame>
+#include <QSettings>
 #include <map>
 
 // ColumnVisibilityDialog implementation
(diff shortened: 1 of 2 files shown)


xusheng6:
Fixes https://github.com/Vector35/debugger/issues/1109

fuzyll (approved):
I don't entirely understand why #1109 is desirable and there's no notes/context, but I agree that this PR should implement everything necessary.

Is the issue that sometimes people want RWX in places and our defaults don't give that as an option? Just wondering if there's maybe a better way to have solved the usability issue than clobbering our defaults with the last selection. Users may still wind up needing to change the access type often, even with this change.

xusheng6:
> I don't entirely understand why #1109 is desirable and there's no notes/context, but I agree that this PR should implement everything necessary.
> 
> Is the issue that sometimes people want RWX in places and our defaults don't give that as an option? Just wondering if there's maybe a better way to have solved the usability issue than clobbering our defaults with the last selection. Users may still wind up needing to change the access type often, even with this change.

Right now the dialog does not save the last selection of the RWX status. For example, if the user is manually tracing back a source of buffer, he would be repeatedly using this dialog to look for writes (W). Right now we do not save the last used selection, which means the dialog would open again with non of the RWX selected, so the user would have to do it every time

Thanks for reviewing and approving!

PR #1151, opened 2026-07-24
Auto-fill the instruction text when adding a TTD bookmark (Fixes #1108)

### LINKED ISSUE #1108: TTD timestamp wdiget should auto-fill the instruciton text when a new timestamp is added

xusheng6:
Prefill the disassembly text of the current instruction as notes when adding a TTD bookmark

### DIFF
--- a/ui/ttdbookmarkwidget.cpp
+++ b/ui/ttdbookmarkwidget.cpp
@@ -27,11 +27,117 @@ static uint64_t PositionSortValue(const TTDPosition& position)
 }
 
 
+// Parses a hex address, with or without the 0x prefix. Returns 0 if the text cannot be parsed
+static uint64_t ParseHexAddress(const QString& text)
+{
+	QString clean = text.trimmed();
+	if (clean.isEmpty())
+		return 0;
+
+	if (clean.startsWith("0x") || clean.startsWith("0X"))
+		clean = clean.mid(2);
+
+	bool ok;
+	uint64_t address = clean.toULongLong(&ok, 16);
+	return ok ? address : 0;
+}
+
+
+static QString FormatTTDPosition(const TTDPosition& position)
+{
+	return QString("%1:%2").arg(position.sequence, 0, 16).arg(position.step, 0, 16);
+}
+
+
+// Parses a "sequence:step" position, both in hex. Returns false if the text cannot be parsed
+static bool ParseTTDPositionText(const QString& text, TTDPosition& result)
+{
+	QStringList parts = text.trimmed().split(':');
+	if (parts.size() != 2)
+		return false;
+
+	bool ok1, ok2;
+	uint64_t sequence = parts[0].toULongLong(&ok1, 16);
+	uint64_t step = parts[1].toULongLong(&ok2, 16);
+	if (!ok1 || !ok2)
+		return false;
+
+	result = TTDPosition(sequence, step);
+	return true;
+}
+
+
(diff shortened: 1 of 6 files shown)


ui/ttdbookmarkwidget.cpp
```diff
@@ -54,6 +160,14 @@ TTDBookmarkEditDialog::TTDBookmarkEditDialog(QWidget* parent, const QString& pos
 	m_noteEdit->setMinimumWidth(400);
 	layout->addRow("Note:", m_noteEdit);
 
+	// Pre-fill the note with the instruction at the view address, so the bookmark is meaningful without the user
```
fuzyll, line 163: Is the logic here correct? I haven't built this and ran it, but if `m_noteEdited` is `!note.isEmpty` (above, line 140), doesn't that mean that deleting a note and leaving it empty will cause it to auto-fill?

If this is intentional/desired, awesome, I just didn't understand that from the comment here. Seemed like an edge-case that wasn't addressed instead.

ui/ttdcallswidget.cpp
```diff
@@ -216,7 +216,7 @@ void TTDCallsQueryWidget::setupUIActions()
 			defaultNote = funcItem->text();
```
fuzyll, line 216: Re: Earlier comment about auto-fill, it looks like this will prevent that. Again, if that's desired, awesome. Just stuck out as a difference.

ui/ttdeventswidget.cpp
```diff
@@ -372,7 +372,7 @@ void TTDEventsQueryWidget::setupUIActions()
 			return;
 		QString posStr = posItem->text();
 
-		TTDBookmarkEditDialog dialog(this, posStr, "", "");
+		TTDBookmarkEditDialog dialog(this, posStr, "", "", m_controller);
```
fuzyll, line 375: I also ran an LLM review on this PR as I had some questions on how certain things worked in the debugger after finishing my review. When I did, it said there is a problem here. I asked it for more information, and it said:

> Exception bookmarks omit the available program counter. The dialog receives an empty view address even though exception rows contain a PC, so these bookmarks neither autofill the instruction nor retain the relevant navigation address.
>
> The bookmark dialog can only generate instruction text when it receives a View Address:
> uint64_t address = ParseHexAddress(m_viewAddressEdit->text());
> if (!data || (address == 0))
>     return "";
> In TTDEventsQueryWidget, exception rows already display the exception’s program counter. For example, `ui/ttdeventswidget.cpp:731` populates the Program Counter column from `event.exception->programCounter`.
> However, the bookmark action always constructs the dialog with an empty view address:
> `TTDBookmarkEditDialog dialog(this, posStr, "", "", m_controller);`
> That third empty string means:
> 1. The dialog cannot disassemble the instruction at the exception PC.
> 2. The generated note remains empty.
> 3. The saved bookmark’s viewAddress remains zero.
> 4. Navigating to that bookmark later changes the TTD position but cannot return the UI to the exception instruction.
> For example, an exception at position `123:4` with PC `0x140001234` creates a bookmark for `123:4`, but discards `0x140001234`.

Looking over this, I think its logic holds?

fuzyll (commented):
Genuinely not sure if I should approve or not. If you go through these and none of them are applicable, let me know and I'll approve. But, I do think there probably need to be _some_ changes, even if they're small.

xusheng6:
This is a new feature so I would come back to it after the stable release

PR #1169, opened 2026-08-12
Fix data race on RspConnector lifetime

### LINKED ISSUE #1074: Data race on m_rspConnector lifetime across Gdb/Corellium/Esreven adapters
## Summary

After #1073 landed a null check on `CorelliumAdapter::BreakInto`, that crash window is narrower but the underlying defect is still present and shared by every adapter that uses `RspConnector`. The null check makes the crash less likely, but it does not stop it from happening.

## Why the null check is insufficient

`DebuggerController::ExecuteAdapterAndWait` deliberately uses two different mutexes (debuggercontroller.cpp:2678-2695). `Pause`/`Quit`/`Detach` take `m_adapterMutex2` while everything else takes `m_adapterMutex`, with a comment that pause must not try to lock the running-operation mutex because it is held by the thread that is currently blocked inside the adapter. In other words `BreakInto` is *designed* to run concurrently with `Go` / `ResponseHandler` on a different thread.

Meanwhile, each adapter's `ResponseHandler` deletes and nulls its own `m_rspConnector` on target-exit (or comparable teardown), with no synchronization on the pointer field. For example, in `CorelliumAdapter::ResponseHandler` at corelliumadapter.cpp:861-864:

```cpp
if (m_rspConnector)
{
    delete m_rspConnector;
    m_rspConnector = nullptr;
}
```

So `BreakInto` reads `m_rspConnector`, and `ResponseHandler` mutates it, with nothing in between. Mutating a field on one thread and reading it from another thread without correct synchronization is a data race. The null check added in #1073 only reduces the window between the load and the call to `SendRaw`; the connector can still be freed in between.

The `EXCEPTION_ACCESS_VIOLATION_READ / 0x50` signature reported in BINARYNINJA-3X is consistent both with a true null deref (offset 0x50 inside `RspConnector`) and with a use-after-free where the freed block was zeroed. The post-#1073 occurrences are most likely the latter.

## Affected adapters

The same raw `RspConnector*` pointer with the same teardown pattern exists in:

- `core/adapters/corelliumadapter.cpp` (h:50 — `RspConnector* m_rspConnector{};`)
- `core/adapters/gdbadapter.cpp`
- `core/adapters/esrevenadapter.cpp`

All three delete the connector inside their own `ResponseHandler` (or comparable teardown) while methods like `BreakInto`, `WriteRegister`, `ReadMemory`, etc. can read it from another thread. Anything that touches `m_rspConnector` outside the operation thread is exposed to the same race.

## Suggested fix

Either:

1. **Mutex**: add a dedicated `std::mutex` per adapter that protects every read and write of `m_rspConnector`. `BreakInto` would acquire it around the `SendRaw` call; the teardown sites would acquire it around delete + null.
2. **`std::atomic<std::shared_ptr<RspConnector>>`**: change the field type, do lock-free atomic loads in callers, atomic-store an empty shared_ptr at teardown. Cleaner but touches every read site.

Either approach should be applied consistently to all three adapters. The shared_ptr form is friendlier if more call sites need protection; the mutex form is a smaller diff if only the teardown paths matter.

## Related

- #1066 (Sentry `BINARYNINJA-3X`) — closed by #1073, but the same crash signature can recur until this is fixed.

xusheng6:
Fixes #1074

`m_rspConnector` is a raw pointer in the Gdb, Corellium and Esreven adapters, with no
synchronization on the read or the delete. A call already in flight can end up using a
freed connector. This is the top debugger crash cluster in Sentry (BINARYNINJA-4E, C1,
E1, E4, ...), which faults inside `mtx_do_lock` while locking `m_socketLock` on a dead
connector.

The existing null checks cannot fix this, because the pointer can be freed between the
check and the use.

## Change

Wrap the field in `AtomicRspConnector`, a mutex-guarded `shared_ptr`. Callers load a
strong reference before use, and teardown just stores nullptr, so the connector is
destroyed only when the last reference drops.

This is 73e6517 from `fix-rspconnector-lifetime-race`, rebased onto dev. One conflict in
`EsrevenAdapter::GetTTDCallsForSymbols`, plus four places where the newer bare
`!m_rspConnector` guards were folded into the `load()`-then-check pattern.

## Testing

`test/repro/` adds an ASan driver that runs `AddBreakpoint` against `Quit` over a fake
RSP server. Same harness and settings both sides:

| build | result |
| --- | --- |
| dev | use-after-free on the first iteration |
| this branch | 25 iterations clean, no ASan report |

The failing stack matches the Sentry reports: freed in `GdbAdapter::Quit`, read from
`AddBreakpoint` -> `TransmitAndReceive` -> `SendPayload` -> `SendRaw`.

### DIFF
--- a/core/adapters/corelliumadapter.cpp
+++ b/core/adapters/corelliumadapter.cpp
@@ -86,7 +86,11 @@ bool CorelliumAdapter::LoadRegisterInfo()
     if (m_isTargetRunning)
         return false;
 
-    const auto xml = this->m_rspConnector->GetXml("target.xml");
+    auto connector = m_rspConnector.load();
+    if (!connector)
+        return false;
+
+    const auto xml = connector->GetXml("target.xml");
 
     pugi::xml_document doc{};
     const auto parse_result = doc.load_string(xml.c_str());
@@ -226,9 +230,10 @@ bool CorelliumAdapter::Connect(const std::string& server, std::uint32_t port)
 		return false;
 	}
 
-    this->m_rspConnector = new RspConnector(this->m_socket);
-    this->m_rspConnector->TransmitAndReceive(RspData("Hg0"));
-    this->m_rspConnector->NegotiateCapabilities(
+    auto connector = std::make_shared<RspConnector>(this->m_socket);
+    m_rspConnector.store(connector);
+    connector->TransmitAndReceive(RspData("Hg0"));
+    connector->NegotiateCapabilities(
             { "swbreak+", "hwbreak+", "qRelocInsn+", "fork-events+", "vfork-events+", "exec-events+",
                          "vContSupported+", "QThreadEvents+", "no-resumed+", "xmlRegisters=i386" } );
     if ( !this->LoadRegisterInfo() )
@@ -242,7 +247,7 @@ bool CorelliumAdapter::Connect(const std::string& server, std::uint32_t port)
     	return false;
     }
 
-    const auto reply = this->m_rspConnector->TransmitAndReceive(RspData("?"));
+    const auto reply = connector->TransmitAndReceive(RspData("?"));
     auto map = RspConnector::PacketToUnorderedMap(reply);
 	this->m_lastActiveThreadId = (uint32_t)map["thread"];
 	this->m_processPid = (uint32_t)map["thread"];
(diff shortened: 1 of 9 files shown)


xusheng6:
@bdash would you please review this? I had a closer look at sentry, and it seems to me that it is still happening

bdash (approved):

PR #1171, opened 2026-08-12
Bounds check RLE runs in DecodeRLE

### LINKED ISSUE #1170: Out-of-bounds read in RspConnector::DecodeRLE on malformed RLE runs
`RspConnector::DecodeRLE` does not bounds check the `'*'` run-length branch:

```cpp
else if (data.m_data[index] == '*')
{
    auto repeat = data.m_data[index + 1] - 29;   // index + 1 may be past the end
    auto last_char = result[result.size() - 1];  // result may be empty -> result[SIZE_MAX]
```

An RLE run is `<char>*<count>`, so it needs both a count byte after the `*` and a
character before it to repeat. Neither is checked.

- A packet whose first byte is `*` leaves `result` empty, so `result[result.size() - 1]`
  indexes `SIZE_MAX`. ASan reports `stack-buffer-overflow ... underflows this variable
  'result'`. `DecodeRLE(RspData("*A"))` and `DecodeRLE(RspData("*"))` both abort.
- A trailing `*` reads `m_data[index + 1]` one past the buffer. `DataBuffer::operator[]`
  is unchecked. ASan does not flag this one because the buffer comes from BN core, which
  is not instrumented, so this part is from reading the code rather than a tool result.

This is reachable from ordinary traffic: RLE is a standard RSP feature and gdbserver uses
it to compress repeated bytes, which is common when reading module maps or zero-filled
memory. The data is entirely stub controlled.

Found by fuzzing the GDB adapter against a live Corellium stub through a mutating proxy.
The path hit was `GetModuleList -> GetRemoteFile -> HostFileIO -> ReceiveRspData -> DecodeRLE`.

Distinct from #1164: that is unhandled exceptions from integer parsing, and #1166 does not
touch `DecodeRLE`. This one is memory corruption, so a catch boundary does not help it.

xusheng6:
Fixes #1170

`DecodeRLE` did not bounds check its `'*'` branch. An RLE run is `<char>*<count>`, so it
needs a count byte after the `'*'` and a character before it to repeat; neither was
checked, and both come from the remote stub.

- `'*'` as the final byte read `m_data[index + 1]` one past the buffer.
- `'*'` before any literal character evaluated `result[result.size() - 1]` on an empty
  string, indexing `SIZE_MAX`.

RLE is a standard RSP feature that gdbserver uses to compress repeated bytes, so this is
reachable from ordinary traffic, not just a hostile stub.

## Change

Reject both cases and log, rather than decoding past the end. A run marker with nothing to
repeat means the packet is malformed. Once #1166 lands this should raise
`DebugProtocolError` like the other malformed-field paths — kept to a log here so the
memory-safety fix can land independently of that PR.

## Testing

Under ASan, before and after:

| input | before | after |
| --- | --- | --- |
| `"*A"` (leading `*`) | abort | ok |
| `"*"` | abort | ok |
| `"AB*"` (trailing `*`) | OOB read, not ASan-visible | ok |
| `"T05;*A"` | ok | ok |

Valid RLE still expands: `"A*!"` -> `"AAAAA"`, `"0* "` -> `"0000"`, and packets with no
`'*'` are unchanged.

Found by fuzzing the GDB adapter against a live Corellium stub through a mutating proxy;
the path hit was `GetModuleList -> GetRemoteFile -> HostFileIO -> ReceiveRspData -> DecodeRLE`.

### DIFF
--- a/core/adapters/rspconnector.cpp
+++ b/core/adapters/rspconnector.cpp
@@ -72,6 +72,21 @@ RspData RspConnector::DecodeRLE(const RspData& data)
             }
             else if (data.m_data[index] == '*')
             {
+                // An RLE run is "<char>*<count>": it repeats the character just emitted, so
+                // both the count byte and that preceding character have to exist. Neither was
+                // checked. A stub sending '*' as the final byte read one past the buffer, and
+                // one sending '*' before any literal character evaluated
+                // result[result.size() - 1] with an empty result, indexing SIZE_MAX.
+                if (index + 1 >= data.m_data.GetLength())
+                {
+                    LogError("Malformed RLE in remote protocol data: run marker with no repeat count");
+                    break;
+                }
+                if (result.empty())
+                {
+                    LogError("Malformed RLE in remote protocol data: run marker with no preceding character");
+                    break;
+                }
                 auto repeat = data.m_data[index + 1] - 29;
                 auto last_char = result[result.size() - 1];
                 for ( auto idx = 0; idx < repeat; idx++ )
--- a/core/adapters/rspconnector.cpp
+++ b/core/adapters/rspconnector.cpp
@@ -31,6 +31,28 @@ RspConnector::RspConnector(Socket* socket) : m_socket(socket) { }
 
 RspConnector::~RspConnector() {}
 
+// Expand one RLE run into result. A run is "<char>*<count>", so it repeats the character
+// already emitted and the count says how many more copies to append. The count is encoded
+// as count + 29 to keep the byte printable: the smallest count byte is ' ' (0x20), which
+// decodes to 3, so "0* " expands to "0000", and the largest is '~' (0x7e). Returns false
+// if there is nothing to repeat or the count is out of range, both of which mean the
+// packet is malformed rather than merely unusual.
+static bool ExpandRleRun(uint8_t countByte, std::string& result)
+{
+	if (result.empty())
+	{
+		LogError("Malformed RLE in remote protocol data: run marker with no preceding character");
+		return false;
+	}
+	if ((countByte < 0x20) || (countByte > 0x7e))
+	{
+		LogError("Malformed RLE in remote protocol data: repeat count out of range");
+		return false;
+	}
+	result.append(countByte - 29, result.back());
+	return true;
+}
+
 RspData RspConnector::BinaryDecode(const RspData& data)
 {
     std::string result{};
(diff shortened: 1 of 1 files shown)


xusheng6:
@bdash can you have a look at this PR and https://github.com/Vector35/debugger/pull/1173? These are found alongside another one you approved earlier when I was fuzzing our RSP implementation

core/adapters/rspconnector.cpp
```diff
@@ -72,6 +72,21 @@ RspData RspConnector::DecodeRLE(const RspData& data)
             }
             else if (data.m_data[index] == '*')
             {
+                // An RLE run is "<char>*<count>": it repeats the character just emitted, so
+                // both the count byte and that preceding character have to exist. Neither was
```
bdash: The first sentence describes what we're checking and why. That seems useful.

The last two thirds of this comment is talking about the previous behavior of this code, which seems a lot less relevant to anyone looking at this code in the future.

core/adapters/rspconnector.cpp
```diff
@@ -72,6 +72,21 @@ RspData RspConnector::DecodeRLE(const RspData& data)
...
+                // checked. A stub sending '*' as the final byte read one past the buffer, and
+                // one sending '*' before any literal character evaluated
+                // result[result.size() - 1] with an empty result, indexing SIZE_MAX.
+                if (index + 1 >= data.m_data.GetLength())
+                {
+                    LogError("Malformed RLE in remote protocol data: run marker with no repeat count");
+                    break;
```
bdash: Breaking seems like it'll return whatever incomplete data has been accumulated so far in `result`. Is that desirable?

core/adapters/rspconnector.cpp
```diff
@@ -72,6 +72,21 @@ RspData RspConnector::DecodeRLE(const RspData& data)
...
+                }
+                if (result.empty())
+                {
+                    LogError("Malformed RLE in remote protocol data: run marker with no preceding character");
+                    break;
+                }
                 auto repeat = data.m_data[index + 1] - 29;
```
bdash: What is this 29 about?

core/adapters/rspconnector.cpp
xusheng6: I shortened the comment by removing the last 2 sentences

core/adapters/rspconnector.cpp
xusheng6: This is about decoding the number of repeats. I added a comment for it

core/adapters/rspconnector.cpp
xusheng6: It now returns {} when decoding error happens.

I also refactored the code and lifted the RLE decoding logic bc I noticed it is also used elsewhere

bdash (approved):

PR #1173, opened 2026-08-12
Invalidate the register cache after a successful register write

### LINKED ISSUE #1172: WriteRegister leaves a stale register cache, so reads return the pre-write value
`WriteRegister` in the Corellium, GDB and esReven adapters writes the register
successfully but does not invalidate the register cache on the success path:

```cpp
const auto reply = this->m_rspConnector->TransmitAndReceive(RspData("P{:02X}={}", ...));
if (reply.m_data[0])
    return true;          // no InvalidateCache() here
```

Only the `G`-packet fallback further down calls `InvalidateCache()`. `ReadRegister`
delegates to `ReadAllRegisters`, which returns `m_regCache` when it is populated, so every
read after a write returns the old value until something else invalidates the cache (a
step or a resume).

The user-visible effect is that editing a register appears to do nothing, and the
debugger keeps computing with the stale value. There is no crash, so this does not show
up in crash reporting.

Confirmed against a live Corellium iPhone 12. The adapter reported the old value for `x0`,
while a raw RSP probe against the same stub showed the write landing:

```
p0                    -> 80ec200002000000     (original)
P0=4142434445464748   -> OK
p0                    -> 4142434445464748     (write landed)
```

So the stub is behaving correctly; the bug is adapter-side caching.

Two further problems in the same two lines:

- `if (reply.m_data[0])` treats any non-empty reply as success. RSP returns `E xx` on
  error, and `'E'` is non-zero, so a failed write is reported as successful.
- `reply.m_data[0]` on an empty reply is an unchecked `DataBuffer` index, the same pattern
  as #1170.

Affects `core/adapters/corelliumadapter.cpp`, `core/adapters/gdbadapter.cpp` and
`core/adapters/esrevenadapter.cpp`, which all carry identical copies of this code.

xusheng6:
Fixes #1172

`WriteRegister` did not invalidate the register cache after a successful `P` write, so
`ReadRegister` kept returning the pre-write value until a step or resume cleared it.
Editing a register looked like it did nothing.

Also check the reply is `"OK"` rather than non-empty: RSP answers `E xx` on failure and
`'E'` is non-zero, so failed writes were reported as successful. An empty reply means `P`
is unsupported and still falls through to the `G` fallback.

Same three lines in the Corellium, GDB and esReven adapters, so all three are fixed.

Verified on a live Corellium device: `WriteRegister("x0")` now reads back what it wrote.

### DIFF
--- a/core/adapters/corelliumadapter.cpp
+++ b/core/adapters/corelliumadapter.cpp
@@ -627,8 +627,11 @@ bool CorelliumAdapter::WriteRegister(const std::string& reg, intx::uint512 value
                                             : uint512ToLittleEndianHex(value, this->m_registerInfo[reg].m_bitSize / 8);
     const auto reply = this->m_rspConnector->TransmitAndReceive(RspData("P{:02X}={}",
                 this->m_registerInfo[reg].m_regNum, newRegString));
-    if (reply.m_data[0])
+    if (reply.AsString() == "OK")
+    {
+        InvalidateCache();
         return true;
+    }
 
     char query{'g'};
     const auto generic_query = this->m_rspConnector->TransmitAndReceive(RspData(&query, sizeof(query)));
--- a/core/adapters/esrevenadapter.cpp
+++ b/core/adapters/esrevenadapter.cpp
@@ -1064,8 +1064,11 @@ bool EsrevenAdapter::WriteRegister(const std::string& reg, intx::uint512 value)
     const auto reply = this->m_rspConnector->TransmitAndReceive(RspData("P{:02X}={}",
                                        this->m_registerInfo[reg].m_regNum, newRegString));
 
-    if (reply.m_data[0])
+    if (reply.AsString() == "OK")
+    {
+        InvalidateCache();
         return true;
+    }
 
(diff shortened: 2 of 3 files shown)


plafosse (approved):
This is approved however i'm concerned about it a bit architecturally. Having to remember to call InvalidateCache at a bunch of different locations across all the adapters seems fragile and error prone.  Could it be moved into `WriteRegister`

xusheng6:
> This is approved however i'm concerned about it a bit architecturally. Having to remember to call InvalidateCache at a bunch of different locations across all the adapters seems fragile and error prone. Could it be moved into `WriteRegister`

it is already inside of `WriteRegister`. The reason you see the code repeated 3 times is that we have 3 copies of similar code, and the reason for that is that we do that to account for the subtle differences and incompatibilities of the specific GDB RSP they each speak

The reason they all work with a GDB is that they tested against it during development. So if we also wish to be compatible with all of them, we need to replicate what GDB does exactly. That is also why I think we would be circulating back to GDB MI bc that feels like the only way to actually "do exactly as GDB"

PR #1174, opened 2026-08-12
X2WinRpcAdapter: remote Windows debugging over FlatBuffers RPC

Weitao-Sun:
`X2WinRpcAdapter` (BN-core) + `x2winstub` (remote Windows stub), talking over
a FlatBuffers RPC protocol, for debugging Windows targets on a separate box.

## What's here

**Supported:** Server-mode and Target-mode connect, launch/attach/detach/quit,
full execution control (go/step into/over/return/break-into), software +
hardware breakpoints (including setting a breakpoint on an already-running
target), memory read/write + memory map, register read/write, thread
list/switch/suspend/resume, module list, call stack, target arch.

**Not supported:** reverse step-over, Time Travel Debugging (TTD).

## Test status

26/27 automated tests pass (`test/x2winrpc_test.py`). See TEST_RESULTS.md for
detail on the one failing test (a bug in the test script itself, not the
adapter/stub) and coverage gaps.

## Known issues (see STATUS.md for details)

1. Conditional breakpoints are slow to evaluate -- generic BN-core code
   (`DebuggerController::ShouldSilentResumeAfterStop()`), not X2Win-specific.
2. `StepReturn()` can pick an unrelated return address if called while the
   thread isn't actually inside a called function's body, on code with no
   real function prologues for `StackWalk64` to unwind.
3. A breakpoint re-armed after `Restart()` can race a caller that resumes
   immediately.
4. Resume after an exception can leave the GUI stuck; doesn't reproduce via
   script, so the adapter/controller-level handling is confirmed correct --
   whatever's wrong is specific to the interactive GUI path.

## Build status

Builds/runs against the remote Windows dev box. Not yet confirmed to pass
this repo's Jenkins CI build.


xusheng6:
Thanks a lot @Weitao-Sun!

PR #1179, opened 2026-08-14
Fix breakpoint removal after debugger exit

### LINKED ISSUE #1175: Crash removing a breakpoint via the toolbar (DbgEngAdapter::RemoveBreakpoint)
Sentry: https://vector35.sentry.io/issues/BINARYNINJA-CR

`EXCEPTION_ACCESS_VIOLATION_READ / 0x0` when removing a breakpoint from the toolbar.
3 occurrences, first 2026-06-22, last 2026-08-06, all on `binaryninja@5.3.9757` (stable),
Windows 10.0.28120 x86_64.

Reported code location is `core/adapters/dbgengadapter.cpp:1276` on `stable_5.3`.

## Stack

```
QToolButton::mouseReleaseEvent            qtoolbutton.cpp:591
QAbstractButtonPrivate::click             qabstractbutton.cpp:365
QAction::activate / triggered
QtPrivate::QSlotObjectBase::call
  - RtlpLowFragHeapAllocFromContext       <unknown>
DebuggerBreakpoints::ContainsAbsolute     debuggerstate.cpp:713
  - std::vector<T>::_Tidy / allocator::deallocate
DebuggerController::DeleteBreakpoint      debuggercontroller.cpp:83
DebuggerBreakpoints::RemoveAbsolute       debuggerstate.cpp:605
  - std::vector<T>::erase                 vector:1793
DbgEngAdapter::RemoveBreakpoint           dbgengadapter.cpp:1276
```

The crashed thread reported 0 frames, so Sentry displayed another thread and the unwind is
partly corrupt — a heap allocation frame and `vector::_Tidy`/`deallocate` are spliced
between the debugger frames. Treat this as the set of frames involved, not their order.

## Hypothesis (unproven)

`vector::erase` + `_Tidy` + `deallocate` appearing next to `ContainsAbsolute` suggests
container/iterator invalidation: `RemoveAbsolute` erasing from the breakpoint vector while
something else still holds an iterator or reference into it. This is inferred from a
partly corrupt stack, not established.

## Repro angles

Windows + DbgEng. Remove a breakpoint **via the toolbar button** (the stack is a
`QToolButton`, not a context menu or the breakpoint pane) at different states:

- while the target is running vs. stopped
- after the target has exited, before teardown
- rapid/double clicks on the remove action
- removing a breakpoint that was never applied, or already removed

## Fix status

Not fixed on `dev`. `dbgengadapter.cpp` has 5 dev-only commits, none touching breakpoint
lifetime (`c8585ad`, `1d522b5`, `a766958`, `e4b8694`, `2ed559b`). `debuggerstate.cpp` got
`723b403` (execution/event model refactor), which could perturb this path but does not
target it. Line numbers above are `stable_5.3`; check dev equivalents before editing.

Not reproduced — investigated from macOS, so this is from the minidump plus reading the
source. DbgEng is Windows-only.

xusheng6:
Avoid backend breakpoint removal after debugger teardown and add regression coverage. Fixes #1175.

A potential follow-up task is identified in https://github.com/Vector35/debugger/issues/1180. However, this is no proof that COM object liftime is causing any issues (this one and beyond) because we are already using an actor model to serialize the accesses to the adapter

### DIFF
--- a/core/debuggerstate.cpp
+++ b/core/debuggerstate.cpp
@@ -683,9 +683,6 @@ bool DebuggerBreakpoints::AddOffset(const ModuleNameAndOffset& address)
 
 bool DebuggerBreakpoints::RemoveAbsolute(uint64_t remoteAddress)
 {
-	if (!m_state->GetAdapter())
-		return false;
-
 	ModuleNameAndOffset info = m_state->GetModules()->AbsoluteAddressToRelative(remoteAddress);
 	auto it = FindBreakpoint(info);
 	if (it == m_breakpoints.end())
@@ -693,7 +690,10 @@ bool DebuggerBreakpoints::RemoveAbsolute(uint64_t remoteAddress)
 
 	m_breakpoints.erase(it);
 	SerializeMetadata();
-	m_state->GetAdapter()->RemoveBreakpoint(remoteAddress);
+
+	if (m_state->GetAdapter() && m_state->IsConnected())
+		m_state->GetAdapter()->RemoveBreakpoint(remoteAddress);
+
 	return true;
 }
 
--- a/test/debugger_test.py
+++ b/test/debugger_test.py
@@ -367,6 +367,23 @@ class DebuggerAPI(unittest.TestCase):
         self.assertEqual(dbg.ip, entry)
         dbg.quit_and_wait()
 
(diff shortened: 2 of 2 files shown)


fuzyll (approved):
This looks good to me.

PR #1182, opened 2026-08-14
Terminate the TTD target on the engine thread

### LINKED ISSUE #1129: DbgEng TTD adapter crashes on stopping debugging when using latest WinDbg
This is due to MS just released a new version of windbg -- it is different from https://github.com/Vector35/debugger/issues/1082

xusheng6:
DbgEng clients are thread-affine, and `ExitDispatch()` is the only call documented as safe from another thread. `Quit()` was calling `TerminateCurrentProcess()` from the requesting thread while the engine thread sat in `DispatchCallbacks()`, which faults inside WinDbg 1.2606's new V8-based `JSProvider.dll`.

`Quit()` now only raises `m_terminateRequested` and wakes the engine loop, which performs the terminate itself.

Reproduced with a harness mirroring the adapter (callbacks, worker thread, engine loop): **5/5 crashes before, 0/10 after** on 1.2606, and 0/3 on 1.2603. WinDbg 1.2402 and 1.2603 never crashed either way, so the pinned version was already a valid mitigation.

Untested path: if the trace is replaying rather than sitting at a break, `Quit()` falls back to `SetInterrupt()`; the harness always stopped at a break.

Fixes #1129

### DIFF
--- a/core/adapters/dbgengadapter.cpp
+++ b/core/adapters/dbgengadapter.cpp
@@ -708,6 +708,14 @@ void DbgEngAdapter::EngineLoop()
 				// WaitForEvent(). The real purpose of this call is to wait until the UI/API initiates another control
 				// operation, which then calls ExitDispatch(), which causes the DispatchCallbacks() to return.
 				m_debugClient->DispatchCallbacks(INFINITE);
+
+				// A DbgEng client belongs to the thread that created it, and ExitDispatch() is the only call
+				// documented as safe to make from another thread. So Quit() only raises this flag and wakes us
+				// up; the terminate itself has to happen here. Doing it from the requesting thread while this
+				// one sits in DispatchCallbacks() faults inside WinDbg's data model JS provider on 1.2606
+				// (#1129).
+				if (m_terminateRequested.exchange(false))
+					TerminateTargetOnEngineThread();
 			}
 			// TODO: add step branch and step backs
 			else if ((execution_status == DEBUG_STATUS_GO) || (execution_status == DEBUG_STATUS_STEP_INTO)
@@ -907,6 +915,15 @@ bool DbgEngAdapter::Detach()
 	m_debugClient->ExitDispatch(reinterpret_cast<PDEBUG_CLIENT>(m_debugClient));
 	return true;
 }
+
+bool DbgEngAdapter::TerminateTargetOnEngineThread()
+{
+	if (!this->m_debugClient)
+		return false;
+
+	return this->m_debugClient->TerminateProcesses() == S_OK;
+}
+
 
 bool DbgEngAdapter::Quit()
 {
--- a/core/adapters/dbgengadapter.cpp
+++ b/core/adapters/dbgengadapter.cpp
@@ -655,6 +655,11 @@ void DbgEngAdapter::EngineLoop()
 	bool outputStateOnStop = settings->Get<bool>("debugger.dbgEngOutputStateOnStop");
 
 	m_lastExecutionStatus = DEBUG_STATUS_NO_DEBUGGEE;
+	// The controller reuses one adapter object across launches, so a terminate request that the previous
+	// session left unserviced (it exited on its own before we picked the request up) would otherwise be
+	// latched here and kill this target. The worker queue cannot deliver a Quit() until the launch
+	// operation completes, which needs the stop event posted below, so no live request can be lost here.
+	m_terminateRequested = false;
 	bool finished = false;
 	while (true)
 	{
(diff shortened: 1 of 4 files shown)


core/adapters/dbgengadapter.h
```diff
@@ -156,6 +161,10 @@ namespace BinaryNinjaDebugger {
 
 		bool m_aboutToBeKilled = false;
 
+		// Raised by Quit() so that EngineLoop() performs the terminate on the thread that owns the
+		// debug client. See TerminateTargetOnEngineThread().
+		std::atomic<bool> m_terminateRequested {false};
```
fuzyll, line 168: I think this might be a mistake? My understanding is that this will _only_ be initialized when you construct the object. But `DebuggerController::CreateDebugAdapter` will _reuse_ the existing object right at the start. You only clear it in one spot, as far as I can tell, and it's not in the TTD stuff below. So, I think it should be possible to wind up with a subsequent launch getting terminated because of a prior run's `true` value.

If I am correct, you don't necessarily need to do something here...you could also just make sure everything clears this in its `Reset`.

core/adapters/dbgengadapter.cpp
```diff
@@ -708,6 +708,14 @@ void DbgEngAdapter::EngineLoop()
 				// WaitForEvent(). The real purpose of this call is to wait until the UI/API initiates another control
 				// operation, which then calls ExitDispatch(), which causes the DispatchCallbacks() to return.
 				m_debugClient->DispatchCallbacks(INFINITE);
+
+				// A DbgEng client belongs to the thread that created it, and ExitDispatch() is the only call
+				// documented as safe to make from another thread. So Quit() only raises this flag and wakes us
+				// up; the terminate itself has to happen here. Doing it from the requesting thread while this
+				// one sits in DispatchCallbacks() faults inside WinDbg's data model JS provider on 1.2606
+				// (#1129).
+				if (m_terminateRequested.exchange(false))
```
fuzyll: Terminating the request here feels wrong? If the termination of the current process fails, which we won't know because we've thrown away the return value of `TerminateTargetOnEngineThread`, won't the controller get stuck because we can't receive the exit event from the engine?

You used to do:

```
if (m_debugClient->TerminateCurrentProcess() != S_OK)
    return false;
```

...down below in the TTD adapter, but since you don't do that anymore, I think the controller might get stuck in `WaitForAdapterStop()`?

I feel like this needs two changes: One to report the error and one to ensure the hang doesn't happen. But, maybe I've misunderstood something. All of this code is new to me, after all.

core/adapters/dbgengttdadapter.cpp
```diff
@@ -303,18 +303,36 @@ bool DbgEngTTDAdapter::SupportFeature(DebugAdapterCapacity feature)
...
+	// Terminating from this thread crashes inside WinDbg's data model JS provider on 1.2606 (#1129):
+	// the engine thread owns the debug client and is parked in DispatchCallbacks(), and DbgEng clients
+	// are thread-affine. Ask that thread to do it and wake it up; ExitDispatch() is the one call that is
+	// documented as safe to make from here.
+	m_terminateRequested = true;
+
+	// If the trace is being replayed rather than sitting at a break, the engine thread is inside
+	// WaitForEvent() where ExitDispatch() will not reach it. SetInterrupt() is safe from any thread and
+	// brings it back to a break, where the request above is picked up.
+	if (m_debugControl && (ExecStatus() != DEBUG_STATUS_BREAK))
```
fuzyll, line 333: Calling `ExecStatus()` here will invoke `GetExecutionStatus()`. Is that thread-affine? The comment above does not mention it at all and [this page](https://learn.microsoft.com/en-us/windows-hardware/drivers/debugger/using-client-objects) suggests it might "fail immediately", whatever that means? None of the `GetExecutionStatus` items on [this page](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/_debugger/) say it can be "called from any thread", either, unlike [`ExitDispatch`](https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/dbgeng/nf-dbgeng-idebugclient-exitdispatch), which does.

fuzyll (commented):
I'm not going to mark this "request changes" because I'm not super confident about my claims or the effect they'll have. But, I'm also not sure we should accept this as-is, either.

core/adapters/dbgengadapter.h
xusheng6: Nice catch, I fixed it in adaddc0223cbdde59b80f0ce65abe81f343b7501

core/adapters/dbgengadapter.cpp
xusheng6: When TerminateCurrentProcess() fails, the controller will not hang, the Quit() is more like requesting a quit, and it does not hurt if it returns true when it should not. I added a LogWarn when the termination fails. I prefer not to add a detection and handling for this case bc when the termination fails, there is not much we can do to improve the situation

core/adapters/dbgengttdadapter.cpp
xusheng6: I am also not exactly sure, but I would say it is fine. ExecStatus() is already called all over the place and not from the main thread, yet it works just fine

fuzyll (approved):
I still think there has to be _something_ better we can do when the debug adapter is stuck, but at least if we're able to warn/error there, it's useful.

Also appears the other issue I raised is now addressed. If we're confident the `ExecStatus()` is fine, I have no remaining issues with this PR. Thanks!

PR #1183, opened 2026-08-17
Document the target memory map being mirrored as segments (#1120)

xusheng6:
Describe the bounded-region-per-mapping behavior added in #1120 in the "The Debugger Memory Region" section: the per-region page permissions, the Memory Map sidebar listing, why it makes Find work during debugging, the refresh-on-stop/rebuild-on-change behavior, and which adapters report a map. Introduce the debugger.useMemoryMapSegments setting that selects between the two models, what each value does, when to turn it off, and that it is sampled at session start.

Also document DebuggerController.memory_map in the how-to guide.

### DIFF
--- a/docs/guide/index.md
+++ b/docs/guide/index.md
@@ -399,6 +399,28 @@ Writing to it will also cause the target's memory to change.
 
 The binary view can be accessed by the ``data`` property of the controller.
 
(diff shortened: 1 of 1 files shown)


xusheng6:
Merging directly since this is docs only and timely

PR #1200, opened 2026-09-16
OpenOCD integration

UvuvDev:
Added OpenOCD Adapter to allow boards/devices to be debugged directly. Inherits from GDB Adapter since OpenOCD uses GDB RSP in the backend. Tested all the basic functionality and it works, besides Step Return (which is not implemented yet for GDB Adapter) and Hardware Breakpoints, which I'm opening a separate issue for.


xusheng6:
I am not against this PR, but this is would become redundant once we have https://github.com/Vector35/debugger/issues/929. The plan is to ship a GDB build on Win/Linux/Mac, and use that to connect to anything talks GDB RSP, instead of trying to make more variants of the very broken GDB RSP class and make it even more fragmented

UvuvDev:
I think I will leave this PR open, convert it to a draft, and then once we get it shipping with the binary I can change this PR so it just spawns OpenOCD.

